PR #2, opened 2016-12-15
email-based user model and google auth

Pomax:
some oddities in this PR due to what manage.py decided to do: the "default inserts" should probably be in a place that doesn't get blasted away when you regenerated migrations "from scratch" because it looks like this branch did exactly that thing...

This definitely needs your eyes @alanmoo, before we merge this in.


app/pulseapi/settings.py
```diff
@@ -19,14 +20,20 @@
...
+# This needs to be a real domain for google auth purposes. As such,
+# you may need to add a "127.0.0.1    test.stuff.com" to your
+# host file, so that google's redirect works. This is the same
+# domain you will be specifying in your Flow credentials, and
+# associated client_secrets.json
+ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', 'test.stuff.com,localhost').split(',')
 
+SESSION_ENGINE = 'django.contrib.sessions.backends.signed_cookies'
+SESSION_COOKIE_AGE = 60 * 60 * 24 * 31
+SECRET_KEY = 'Oh my god I love cake so much holy shit how amazing is cake; like, seriously?'
```
alanmoo: lol

app/generate_client_secrets.py
```diff
@@ -0,0 +1,24 @@
...
+    'client_secret': os.getenv('client_secret', 'no secret set'),
+    'redirect_uris': os.getenv('redirect_uris', 'http://localhost:0').split(','),
+    'auth_uri': os.getenv('auth_uri', 'http://localhost:0'),
+    'token_uri': os.getenv('token_uri', 'http://localhost:0')
+  }
+}
+
+data = json.dumps(client_secrets)
+
+text_file = open("client_secrets.json", "w")
```
alanmoo: This should create the file if it doesn't already exist.

alanmoo: To get `migrate` to work, I had to paste in the `web` object and all its keys (with empty strings) to a manually generated `client_secrets.json`.

alanmoo: We need this rigamarole of env var -> JSON because of the way Heroku works, right?

Pomax: not strictly heroku, but yes: we don't want to check in a client_secrets.json, and the auth library does not read in values from env, so we need a little script that can generate the file the library wants, based on variables in the active environment

Pomax:
Note: I renamed the `hosts` binding to `test.example.com` instead.

Pomax:
@alanmoo one thing that is not "going right" is that tags are not getting split on commas before passing it on to the serializer - the html form I wrote for testing does not do a `<select>` for tags with option to write your own, so we probably want to figure out what to do here: grab the list of tags and issues from `/tags.json` and `/issues.json`, and use some "intelligent" dropdown that lets you pick-and-specify-if-you-need-more?

alanmoo:
I'd think issues can be a `<select multiple>`,  but tags are separatable if passed as multiple strings (`["foo","bar"]`), so I was thinking the form should run a split and trim before posting. How's that sound?

Pomax:
sounds fair, I'll update the `public/index.html` test page. One other thing: right now the `submitted_by` needs to be passed by the form, but that seems very wrong: that should be a value that gets autogenerated based on the logged in user at the time of POST-ing by the API server, so you can't post "as someone else". Thoughts?

Pomax:
hm, no matter how I pack the tags in `FormData`, the server keeps seeing a full string, so I might need to unpack them on the server instead of expecing FormData to get unpacked "properly"

Pomax:
@alanmoo right, I've tried about 12 different ways to get "an array" into the entries post route, and nothing works. It keeps ending up either the string representation , or breaking the POST because somehow I'm trying to send an illegal value. I can only conclude that there is some kind of Django magic I'm missing but I'm getting way too pissed off with Django's magic atm so I'm calling it a day =)

PR #6, opened 2016-12-20
Creator and publisher

alanmoo:


.vscode/launch.json
```diff
@@ -0,0 +1,168 @@
+{
```
Pomax: should this be in .gitignore?

.vscode/settings.json
```diff
@@ -0,0 +1,4 @@
+// Place your settings in this file to overwrite default and user settings.
```
Pomax: should this be in .gitignore?

pulseapi/creators/migrations/0002_auto_20161219_2146.py
```diff
@@ -0,0 +1,27 @@
+# -*- coding: utf-8 -*-
```
Pomax: as a new model, these two migrations can be collapsed (delete 001 and 002, then rerun makemigrations)

pulseapi/creators/models.py
```diff
@@ -0,0 +1,32 @@
...
+from pulseapi.users.models import EmailUser
+
+class CreatorQuerySet(models.query.QuerySet):
+    """
+    A queryset for creators which returns all creators by name
+    """
+
+    def public(self):
+        """
+        Returns all creators to start
```
Pomax: what does this mean? this function definitely doesn't do what this comment says, it's just a reflection function for get a reference to an object (to which you would already have a reference, so not sure what this function is for. What uses this?)

pulseapi/creators/tests.py
```diff
@@ -0,0 +1,3 @@
+from django.test import TestCase
```
Pomax: do we want to keep this since we don't have any tests atm?

pulseapi/entries/models.py
```diff
@@ -56,3 +61,4 @@ class Meta:
 
     def __str__(self):
         return str(self.title)
+
```
Pomax: missing new line

pulseapi/users/admin.py
```diff
@@ -1,5 +1,15 @@
 from django.contrib import admin
 
+from pulseapi.entries.models import Entry
 from .models import EmailUser
 
-admin.site.register(EmailUser)
+class EntryInline(admin.StackedInline):
```
Pomax: what does this do? I don't see it referenced anywhere (might need some documentation)

.vscode/launch.json
alanmoo: I was torn on this- I figured it should be committed because it's useful if you happen to be using that editor (or if I switch computers) but I do see the opposite argument as well.

.vscode/settings.json
alanmoo: Same as above

pulseapi/creators/tests.py
alanmoo: It was generated automatically with `manage.py startapp`, I'm fine leaving this as a stub for future tests

PR #8, opened 2016-12-21
DRF tracking for entries endpoints

alanmoo:
Simple but easy

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -8,6 +8,7 @@ from rest_framework.decorators import detail_route
 from rest_framework.generics import ListCreateAPIView, RetrieveAPIView
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
+from rest_framework_tracking.mixins import LoggingMixin
 
 from pulseapi.entries.models import Entry
 from pulseapi.entries.serializers import EntrySerializer
@@ -67,7 +68,7 @@ class EntryCustomFilter(filters.FilterSet):
         fields = ['tags', 'issues',]
 
 
-class EntryView(RetrieveAPIView):
+class EntryView(LoggingMixin, RetrieveAPIView):
     """
     A view to retrieve individual entries
     """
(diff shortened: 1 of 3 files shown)


Pomax:
very nice! though, is there a way to specify the config that says what to log and to where?

alanmoo:
As in what fields? It's a pretty straightforward mixin that attaches to view classes, doesn't seem like there's much of an API to configure it. What's an example of something you're looking for?

Pomax:
a way to feed that log into a database, or a file, etc. for one - a massive console log makes log parsing hard *and* console error debugging hard =)

alanmoo:
This is all accessible in the admin database; Are you looking for more than that?
image

Pomax:
Aggregates, mostly. individual entries like these are going to supersize the database incredibly fast if we don't put something in place to perform the aggregation. 

I'm also worried about the IP logging here. With logs done this way, we now have route access logs bound to IP and timestamped, that can be turned over to authorities. A preferable logging system would store "hashed IP to views" information only (so we can't track actions to specific users even if compelled) as well as "timestamps for view operations" so we can see which entry is popular when, but not a breakdown as to which IPs are responsible for those views at which time.

Should we worry about that right now? As a dev I would say no, but as a user, I would say absolutely. This is something worth doing right from the get go. @xmatthewx has thought be put into the level of statistics gathering that is necessary, and whether we can do that in a way that does not invade our users' privacy, or expose them to potential abuse by governmental order?

alanmoo:
Great points, though it really starts to push me towards [Piwik](https://piwik.org), so we don't have to reimplement things. Their docs seem to cover everything you're suggesting. http://piwik.org/docs/privacy/

alanmoo:
@simonwex do you have thoughts about this?

alanmoo:
Linked to https://github.com/mozilla/network-pulse/issues/251

alanmoo:
As per https://github.com/mozilla/network-pulse/issues/251 I think I'm going to head in a different direction than this PR.

PR #9, opened 2016-12-22
Test suite

Pomax:
forming a PR because it lets me actually conveniently comment on tests =)


pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class TestEntryView(TestCase):
+    def setUp(self):
+        self.entries = [EntryFactory() for i in range(2)]
+        for entry in self.entries:
+            entry.save()
+
+    def test_get_single_entry_data(self):
+        """
```
Pomax: we probably want to test for "by title" with two same-titled entries too

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+        Check if we can get a single entry by its `id`
+        """
+
+        id = self.entries[0].id
+        response = self.client.get(reverse('entry', kwargs={'pk': id}))
+        self.assertEqual(response.status_code, 200)
+
+    def test_post_minimum_entry(self):
+        #Entry with minimum amount of content (title, description?)
```
Pomax: do we want negative tests, too, so that there's no way to regress to a model where title+description are no longer necessary?

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+    def test_post_entry_with_mixed_creators(self):
+        #Entry with some existing creators, some new creators
+
+    def test_get_entries_list(self):
+        #Get /entries endpoint
+
+    def test_get_tag_list(self):
+        #maybe for another view. But tags list should have multiple tags, some with multiple entries associated
+        #do we need an endpoint for /tags/{tagname} that has the associated entries? or does /entries/?tag=foo work?
+
```
Pomax: we probably want an issues list too, since those'll have to be API exposed for 3rd parties (so they can generate an issue dropdown)

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+        #test searching entries
+
+    def test_entries_tags(self):
+        #test filtering entries by tag
+
+    def test_entries_issue(self):
+        #test filtering entires by issue
+
+    def test_post_entry_new_issue(self):
+        #posting an entry with a new Issue should result in an error. Permission denied?
```
Pomax: probably not permission denied, but maybe a 409 "please rephrase yourself" with a specific JSON response that explains the problem is a non-existent issue?

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+
+    def test_check_for_issues(self):
+        #make sure 5 issues are in the database. Does testing use a "real" database?
+
+    def test_post_authentication_requirement(self):
+        #uhhhh how do we do this?
```
Pomax: hmm... can we fake request.session? these might need to be unit tests on the `entries` and `users` views.py instead though

requirements.txt
```diff
@@ -3,9 +3,15 @@ Django==1.10.3
 django-cors-middleware==1.3.1
 django-filter==1.0.0
 djangorestframework==3.5.3
+factory-boy==2.8.1
+Faker==0.7.7
 google-api-python-client==1.5.5
 httplib2==0.9.2
 oauth2client==4.0.0
 oauthlib==2.0.0
+pyasn1==0.1.9
```
Pomax: what keeps adding pyasn.... O_O

(from what I can tell nothing actually *needs* it, it just gets added for... convenience?)

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,67 @@
...
+        #test filtering entries by tag
+
+    def test_entries_issue(self):
+        #test filtering entires by issue
+
+    def test_post_entry_new_issue(self):
+        #posting an entry with a new Issue should result in an error. Permission denied?
+
+    def test_check_for_issues(self):
+        #make sure 5 issues are in the database. Does testing use a "real" database?
```
Pomax: it uses the real database, but we may need to verify it uses django.test so that tests on the db happen as transactions (i.e. properly independent)

pulseapi/entries/test_views.py
alanmoo: Make two entries with the same title and see if we get two entries back.

pulseapi/entries/test_views.py
alanmoo: Don't authenticate, then try to post.

Pomax: https://httpstatuses.com/422 looks pretty appropriate

alanmoo:
I don't think I can request you to review this since you opened it @pomax but I think it's ready.

alanmoo:
(Though I realize I didn't implement some of the things from your comments. Digging into those now)

pulseapi/entries/test_views.py
alanmoo: Not sure I follow this. To rephrase my original comment though, I was hoping the migrations from the original issue generation would be in there. It looks like they are, as writing this test was trivial.

requirements.txt
alanmoo:
```
Collecting pyasn1>=0.1.7 (from oauth2client==4.0.0->-r requirements.txt (line 10))
  Downloading pyasn1-0.1.9-py2.py3-none-any.whl
Collecting pyasn1-modules>=0.0.5 (from oauth2client==4.0.0->-r requirements.txt (line 10))
  Downloading pyasn1_modules-0.0.8-py2.py3-none-any.whl
```

pulseapi/entries/test_views.py
alanmoo: I've ended up attempting to get a nonce as an unauthenticated user. That gives a 404, which is a passing test.

pulseapi/creators/models.py
```diff
@@ -15,9 +15,12 @@ def public(self):
         """
         return self
 
+    def slug(self, slug):
+        return self.filter(name=slug)
+
 class Creator(models.Model):
     """
-    A pulse entry
+    Person responsible for doing the thing an entry is about
```
Pomax: "Person recognized as the creator of the thing an entry links out to"?

pulseapi/creators/tests.py
```diff
@@ -1,3 +1,64 @@
...
+            'creators': ['Pomax','Alan'],
+            'interest': 'interest field',
+            'get_involved': 'get involved text field',
+            'get_involved_url': 'http://example.com/getinvolved',
+            'thumbnail_url': 'http://example.com/',
+            'content_url': 'http://example.com/',
+            'internal_notes': 'Some internal notes'
+        })
+        creatorList = json.loads(str(self.client.get('/creators/?search=A').content, 'utf-8'))
+        self.assertEqual(creatorList, ['Alan'])
```
Pomax: why does github even care about this?

pulseapi/entries/serializers.py
```diff
@@ -39,6 +42,9 @@ class EntrySerializer(serializers.ModelSerializer):
     issues = serializers.SlugRelatedField(many=True,
                                           slug_field='name',
                                           queryset=Issue.objects)
+    creators = CreatableSlugRelatedField(many=True,
+                                            slug_field='name',
```
Pomax: nit: indent looks off, no idea if PEP8 has opinions here

pulseapi/creators/tests.py
alanmoo: Truly a question I've wondered for a while now

pulseapi/entries/test_models.py
```diff
@@ -0,0 +1,68 @@
...
+faker = FakerFactory.create()
+
+
+class EntryFactory(factory.DjangoModelFactory):
+
+    # def __init__(self, tags=['foo','bar']):
+    #     self.tags = tags
+    #     print(self.tags)
+
+    title = factory.LazyAttribute(lambda o: 'title '+' '.join(faker.words(nb=1)))
```
Pomax: we probably want a few more than just one word for title, especially since we restricted it to only 140 characters

pulseapi/entries/test_models.py
```diff
@@ -0,0 +1,68 @@
...
+from pulseapi.entries.models import Entry
+from pulseapi.users.test_models import EmailUserFactory
+
+
+faker = FakerFactory.create()
+
+
+class EntryFactory(factory.DjangoModelFactory):
+
+    # def __init__(self, tags=['foo','bar']):
```
Pomax: there's a bit of commenting in this file that we can probably clean up.

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,275 @@
...
+from django.test import TestCase
+from django.test import Client
+
+from pulseapi.users.test_models import EmailUserFactory
+from pulseapi.entries.test_models import EntryFactory
+
+
+class TestEntryView(TestCase):
+    def setUp(self):
+        self.entries = [EntryFactory() for i in range(2)]
```
Pomax: Hmmmm this looks identical to the code in creators - worth DRYing out?

pulseapi/entries/test_views.py
```diff
@@ -0,0 +1,275 @@
...
+        password = "password1234"
+        user = EmailUserFactory(email=email, password=password, name="test user")
+        user.save()
+
+        self.client = Client()
+        self.client.force_login(user);
+
+        values = json.loads(str(self.client.get('/nonce/').content, 'utf-8'))
+        # Set up with some curated data for all tests to use
+        postresponse = self.client.post('/entries/', data={
```
Pomax: we should probably capture this `data` object, and then for each test copy it and modify as needed.

pulseapi/users/test_views.py
```diff
@@ -0,0 +1,43 @@
...
+from pulseapi.users.test_models import EmailUserFactory
+
+
+class TestUserView(TestCase):
+    def setUp(self):
+        self.users = [EmailUserFactory() for i in range(2)]
+        for user in self.users:
+            user.save()
+
+    # def test_list_users_returns_user_data(self):
```
Pomax: are these bad tests?

requirements.txt
Pomax: oauth~~~

pulseapi/users/test_views.py
alanmoo: I'm not sure where these came from, but yeah I don't think they're useful. Removing.

alanmoo:
Ok, I've updated this branch @Pomax. Take another look?

Pomax:
R+, lets merge this in

Issue #12, opened 2017-01-09
Add connection between entries and user favorites


alanmoo:
So that we can preserve authenticated user favorites.

Pomax:
We should be able to set up a `{user, entry, timestamp}` relation. Even if no built in fields allow for that, there's always the option of making a three-field model and tracking that as its own genuinely independent thing... although I suspect django already has things in place for us to not need to go quite that far.

PR #14, opened 2017-01-13
User bookmarking

Pomax:
- `/entries` and `/entries/1234` show entries/a single entry with their total fave count as field `favorite_count`, and a field `is_favorite` that is `true` for a user that is a) logged in, and b) has favorited this entry. In any other case, it will be `false`.
- `/entries/123/favorite` toggles a fave/not-fave value for entry `123` for the currently logged in user. If a user is not logged in, they will receive a 403

fixes https://github.com/mozilla/network-pulse-api/issues/12
fixes https://github.com/mozilla/network-pulse-api/issues/30


pulseapi/users/test_views.py
```diff
@@ -0,0 +1,12 @@
+from django.core.urlresolvers import reverse
```
alanmoo: Is this file old?

pulseapi/users/serializers.py
```diff
@@ -1,7 +1,24 @@
...
+    UserFavorites,
+)
+
+
+class UserFavoritesSerializer(serializers.ModelSerializer):
+    """
+    Serializes a {user,entry,when} fave.
+    """
+
+    user = '__all__'
```
alanmoo: What's this do?

pulseapi/users/admin.py
```diff
@@ -2,19 +2,35 @@
...
+    readonly_fields = ('entries','favorites')
+
+    # this allows us to create/edit/delete/etc favorites:
+    inlines = [ UserFavoritesInline ]
+
     def entries(self, instance):
         """
         Show all entries as a string of titles. In the future we should make them links.
         """
+        print('entries instance')
```
alanmoo: Remove this logging

pulseapi/entries/views.py
```diff
@@ -4,13 +4,54 @@
...
+from rest_framework.decorators import detail_route, api_view
 from rest_framework.generics import ListCreateAPIView, RetrieveAPIView
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
 from pulseapi.entries.models import Entry
 from pulseapi.entries.serializers import EntrySerializer
+from pulseapi.users.models import UserFavorites
+
+# FIXME: TODO: Remove the handling for GET, so that only PUT works
```
alanmoo: Is the GET there just so it's easier to test?

pulseapi/entries/views.py
```diff
@@ -4,13 +4,54 @@
...
+@api_view(['GET','PUT'])
+def toggle_favorite(request, entryid):
```
alanmoo: Just so I understand what's going on here because I've seen this before, can you explain line 17 and 18 from a django standpoint? Why is this a function and not a view class? Is 17 the indicator that it's a generic GET/PUT view which does nothing special?

alanmoo:
Can you add some tests?

A little bit of clean up (the FIXME's), and removing logging, but generally this looks good. I like the feel of the endpoint behaviors.

pulseapi/users/test_views.py
Pomax: I think so, I'll remove it.

pulseapi/users/serializers.py
Pomax: good question @gideonthomas do you remember why we needed this? If so, we can document it in the source

pulseapi/entries/views.py
Pomax: correct. The "real" use would be PUT only, but for testing a GET is way easier to do without writing an XHR test page

pulseapi/entries/views.py
Pomax: line 17 says "only expose this `def` when the http verb is GET or PUT (http verb whitelist)

line 18 is a function that services a view (see urls.py) as a general function rather than a class entry, as faving an entry is technically something "outside" of entries and users.

pulseapi/users/serializers.py
gideonthomas: Actually, I don't know what this is :P I thought all serializer fields should be instances. Are you sure this is working?

Pomax: `__all__` normally exposes all public methods for use in `import ... from ... ` statements including those prefixed with underscores (usually 'hidden') but in this case it doesn't look like it actually does anything, so I'll remove it.

pulseapi/creators/tests.py
```diff
@@ -2,26 +2,63 @@
 
 from django.test import TestCase
 from django.test import Client
-from pulseapi.tests import PulseTestCase
```
Pomax: Did my PR reset this to older code?

pulseapi/entries/test_views.py
```diff
@@ -1,9 +1,8 @@
 import json
 
-from django.core.urlresolvers import reverse
```
Pomax: hm, this looks to have old test code, too.

pulseapi/entries/views.py
```diff
@@ -21,8 +56,8 @@ def post_validate(request):
     nonce = False
 
     if request.data:
-        csrf_token = request.data['csrfmiddlewaretoken']
-        nonce = request.data['nonce']
+        csrf_token = request.data.get('csrfmiddlewaretoken', False)
```
Pomax: this was throwing key errors so I turned it into `get` rather than direct access to get a fallback value.

pulseapi/entries/serializers.py
```diff
@@ -50,6 +57,35 @@ class EntrySerializer(serializers.ModelSerializer):
...
+
+    def get_bookmark_count(self, instance):
+        """
+        Get the total number of bookmarks this entry received
+        """
+        return instance.bookmarked_by.count()
+
+    is_bookmarked = serializers.SerializerMethodField()
+
+    def get_is_bookmarked(self, instance):
```
alanmoo: This is the kind of magic that bothers me. I understand it, but it still feels weird.

pulseapi/entries/views.py
```diff
@@ -4,13 +4,48 @@
...
+from rest_framework.generics import ListCreateAPIView, RetrieveAPIView, ListAPIView
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 
 from pulseapi.entries.models import Entry
 from pulseapi.entries.serializers import EntrySerializer
+from pulseapi.users.models import UserBookmarks
+
+@api_view(['PUT'])
+def toggle_bookmark(request, entryid):
```
alanmoo: We're sticking with the toggling behavior?

pulseapi/creators/migrations/0001_initial.py
```diff
@@ -21,7 +18,6 @@ class Migration(migrations.Migration):
             fields=[
                 ('id', models.AutoField(auto_created=True, primary_key=True, serialize=False, verbose_name='ID')),
                 ('name', models.CharField(max_length=140)),
-                ('user', models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, related_name='as_creator', to=settings.AUTH_USER_MODEL)),
```
alanmoo: I know I probably shouldn't be reading into the migrations files too much but why is this being removed?

Should we merge these and then follow up with an immediate rebuild of migrations?

pulseapi/entries/serializers.py
Pomax: I hate Django for this kind of thing.

pulseapi/entries/views.py
Pomax: After thinking about it some more, entries already say whether they are bookmarked or not when you call `/entries` so for regular operation we don't need a "toggle yes or no", the only time we *would* need that is for bulk bookmark syncing, which we should probably do on its own dedicated route so a full set of bookmarks can be sent over in a single call.

alanmoo: That makes sense. There's then a little bit of manual state management between the front and back end, but it seems reasonable.

pulseapi/creators/migrations/0001_initial.py
Pomax: that would be my suggestion too, yes. It's time to wipe the compound migrations we've been accumulating (except for issues, which is one step and has data) and rebuild everything as single step migrations.

alanmoo: Cool. Wanna wrap that into this PR so we don't have a weird state on master?

PR #15, opened 2017-01-13
Page to post an array of entries for migration

alanmoo:


public/migrate.html
```diff
@@ -0,0 +1,161 @@
...
+      var data = evt.target.response;
+      try {
+        data = JSON.parse(data);
+        currentUser = data.user;
+        currentNonce = data.nonce;
+        currentCSRF = data.csrf_token;
+
+        // when we have a nonce, set up actual form handling
+        var btn = document.querySelector("#submit");
+        btn.onclick = submitForm
```
Pomax: tiny user improvement: when you click "submit", grey out the submit button again, so you can't restart the submission process?

Pomax:
We probably want to streamline this a little bit so we can throw away old style code:

- enable the button when text is put into the textarea (*and* `JSON.parse` doesn't error on that input =)
- when clicking submit, convert text into an `entries` array
- start to iterate through the array:
  - doNextEntry() => entry = entries.pop(), get nonce
  - getNonce(entry) => handleNonceRequest
  - handleNonceRequest(entry, data) => prepare data payload based on entry + data.nonce etc
  - postData(data) => xhr with 'done' callback to doNextEntry

public/migrate.html
```diff
@@ -0,0 +1,144 @@
...
+      entries = JSON.parse(document.getElementById("entryArray").value);
+      getNonce(postAnEntry);
+    };
+
+    // don't let the browser submit this form. We'll handle it.
+    document.forms.entries.onsubmit = () => false;
+    document.querySelector("#submit").onclick = submitForm;
+    var textBox = document.getElementById("entryArray");
+    var submitBtn = document.querySelector("#submit");
+    textBox.onkeyup = (e) => {
```
Pomax: will this trigger on a mouse right-click + paste?

alanmoo: No. What handler would you suggest? `onchange` seemed to depend on focus, so that wasn't great either.

Pomax:
R+!

PR #17, opened 2017-01-13
Make this work on heroku

alanmoo:


requirements.txt
```diff
@@ -9,6 +9,7 @@ google-api-python-client==1.5.5
 httplib2==0.9.2
 oauth2client==4.0.0
 oauthlib==2.0.0
+psycopg2==2.6.2
```
Pomax: I don't see any changes to the code above that adds in this dependency, where did it come from?

Pomax:
This looks R+, but why is there a new dependency in the requirements.txt?

alanmoo:
I believe heroku was not happy without that dependency

Pomax:
Is there a log that explains why? If it passes now, feel free to merge it in, I'm just not familiar with django enough yet to go "oh yeah of course, dependency X, obviously we need that" =P

alanmoo:
There's probably a long somewhere deep in heroku, but I don't know it's worth the effort to find it. In any case, I haven't added any dependencies recently, so *something* wanted it there.

Issue #16, opened 2017-01-13
restrict logins to moz-org users only


Pomax:
restrict the oauth callback to check whether an account is associated with the mozilla organization. If so, continue login callback, if not, do not allow login.

PR #19, opened 2017-01-15
Email lockdown

Pomax:
fixes #https://github.com/mozilla/network-pulse-api/issues/16

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -21,8 +21,8 @@ def post_validate(request):
     nonce = False
 
     if request.data:
-        csrf_token = request.data['csrfmiddlewaretoken']
-        nonce = request.data['nonce']
+        csrf_token = request.data.get('csrfmiddlewaretoken', False)
+        nonce = request.data.get('nonce', False)
     else:
         csrf_token = request.POST.get('csrfmiddlewaretoken', False)
         nonce = request.POST.get('nonce', False)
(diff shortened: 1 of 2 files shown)


Pomax:
Tested this with my `mozillafoundation.org` account, which works, and my `gmail.com` and `hotmail.com` accounts, which get blocked. Admin login not being a moz login is essentially irrelevant as it uses its own login portal on `/admin`

alanmoo:
While this technically works, I'm concerned about the silent failure if non-Mozillians attempt to log in. I feel like we need some way to convey "you're not logged in but that's expected".

Pomax:
fair enough - rather than a redirect to `/` we can redirect to `/login-failure` so that we can serve up some content with a 403?

alanmoo:
Or maybe a query param so that users can still get to the same page when merged with #25?

Pomax:
oh that's a good idea!

Pomax:
Updated the PR so that it now does a callback with a `loggedin=...` that is either True or False depending on whether authentication succeeded or not, and will either redirect the user to `/` with that argument if no original_url is provided, or redirects to the original_url with that query pair added (with a toggle to make sure we use `?loggedin=...` or `&loggedin=...` depending on whether the original url already had query args)

pulseapi/entries/views.py
```diff
@@ -21,8 +21,8 @@ def post_validate(request):
     nonce = False
 
     if request.data:
-        csrf_token = request.data['csrfmiddlewaretoken']
-        nonce = request.data['nonce']
+        csrf_token = request.data.get('csrfmiddlewaretoken', False)
```
Pomax: yeah, it's a property of Python's dictionaries. You can either do direct access with array notation, or you can use `get` with a fallback, https://docs.python.org/2/library/stdtypes.html#dict.get

Issue #24, opened 2017-01-16
add a redirect argument for login from "some other site" 


Pomax:
Mostly so that post-auth we can redirect the browser back to the URL the user was on prior to login.

PR #25, opened 2017-01-16
This adds a ./login route that can take ?original_url=....

Pomax:
fixes https://github.com/mozilla/network-pulse-api/issues/24

This adds a `/login` route that clients/browsers can be redirected to, which automatically kicks off google oauth. Upon completion, the user is either redirected to `/`, or if they called `/login?original_url=.....` they will be redirected to the passed url. This allows users to log in through 3rd party web clients and be sent back to the page they were on when they went to log in.

### DIFF
--- a/pulseapi/users/urls.py
+++ b/pulseapi/users/urls.py
@@ -4,6 +4,7 @@ from . import views
 
 urlpatterns = [
     url(r'^$', views.index, name='index'),
+    url(r'^login$', views.start_auth, name='login'),
     url(r'^logout', views.force_logout, name='logout'),
     url(r'^oauth2callback', views.callback, name='oauthcallback'),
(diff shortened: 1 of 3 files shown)


Pomax:
@alanmoo I was never able to do that, I had to log out and be in a browser without active google sessions. Did you try this in a private browser tab?

alanmoo:
@pomax Sorry, you're right- looks like my browser was only signed in to one Google account when I first tested this.

Pomax:
google auth is both super convenient, and kind of not, that way =P

I'm not sure what the github UI is telling me about this PR now... does it count as r+?

alanmoo:
I don't know how it handles multiple reviewers, I'm guessing it wants everyone to approve. But as far as I'm concerned it's R+

Issue #22, opened 2017-01-16
Most of the Entry's fields should be optional


mmmavis:
## Required

- title
- content_url
- published_by (correct me if I'm worng - we can pull this directly from the logged in user profile and there's no need to have it on the submission form.)

## Optional

- description
- thumbnail_url
- tags
- get_involved
- get_involved_url
- interest
- internal_notes
- featured
- issues
- creators

/cc @kristinashu @xmatthewx

mmmavis:
Sorry just checked `models.py` and seems like there are more fields we should make them optional. I'll update the ticket description soon.

mmmavis:
Ticket title and description have been updated!

kristinashu:
Looks right to me!

PR #31, opened 2017-01-17
Simplify required fields

alanmoo:
Closes https://github.com/mozilla/network-pulse-api/issues/22

How are migrations handled when a server is already live? Do I need to just `makemigration` starting from a database that's identical to the one that's on the server and then commit the migration files?

### DIFF
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -23,17 +23,17 @@ class Entry(models.Model):
     A pulse entry
     """
     title = models.CharField(max_length=140)
-    description = models.CharField(max_length=300)
-    content_url = models.URLField()
-    thumbnail_url = models.URLField()
+    description = models.CharField(max_length=300, blank=True)
+    content_url = models.URLField(blank=True)
+    thumbnail_url = models.URLField(blank=True)
     tags = models.ManyToManyField(
         Tag,
         related_name='entries',
         blank=True,
     )
-    get_involved = models.CharField(max_length=300)
-    get_involved_url = models.URLField()
-    interest = models.CharField(max_length=300)
+    get_involved = models.CharField(max_length=300, blank=True)
+    get_involved_url = models.URLField(blank=True)
+    interest = models.CharField(max_length=300, blank=True)
     featured = models.BooleanField()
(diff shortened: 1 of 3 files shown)


pulseapi/entries/models.py
```diff
@@ -23,17 +23,17 @@ class Entry(models.Model):
     A pulse entry
     """
     title = models.CharField(max_length=140)
-    description = models.CharField(max_length=300)
-    content_url = models.URLField()
-    thumbnail_url = models.URLField()
+    description = models.CharField(max_length=300, blank=True)
+    content_url = models.URLField(blank=True)
```
Pomax: Are we sure about this one? Are people supposed to be able to post entries without a URL to point to the thing the entry was supposed to be for?

alanmoo: You're right, I goofed. Fixed now.

pulseapi/entries/test_views.py
```diff
@@ -19,7 +19,7 @@ def test_post_minimum_entry(self):
         """
         Test posting an entry with minimum amount of content
         """
-        payload = self.generatePostPayload(data={'title':'title test_post_minimum_entry'})
+        payload = self.generatePostPayload(data={'title':'title test_post_minimum_entry', 'tags':''})
```
Pomax: We probably also want to make sure we have full test coverage for required fields now. https://github.com/mozilla/network-pulse-api/blob/fb8a57cff6fcd027bf9eb9766a71bd1ae5d7733f/pulseapi/entries/test_views.py#L38 tests for title, but with `content_url` also being required, we want to make sure that's a 400 as well?

alanmoo: Added.

pulseapi/entries/test_views.py
```diff
@@ -36,23 +36,18 @@ def test_post_duplicate_title(self):
         self.assertEqual(len(entriesJson), 4)
 
     def test_post_empty_title(self):
-        """Make sure entries require a title"""
+        """Make sure entries require a title and content_url"""
```
Pomax: can we split this up into two tests?

Issue #26, opened 2017-01-17
Entry model should have a timestamp field


mmmavis:

Pomax:
what does this timestamp represent? datetime the entry got posted?

mmmavis:
I believe so (cc @kristinashu for confirmation).

kristinashu:
Yes, same as on production site. Example "Added Dec 12, 2016" https://mzl.la/pulse?id=p12122016154012

mmmavis:
I assume this won't be hard to implement so I've labelled this ticket as "MVP for Pulse Launch". If it's not a simple task we can wait postpone this ticket till later.

PR #32, opened 2017-01-18
Add timestamp to entries

alanmoo:
Closes https://github.com/mozilla/network-pulse-api/issues/26

Again, just need to understand what to do about migrations with this


Pomax:
For migrations I'd say let's pick a moment when all of this has landed, and then we wipe all the migrations (except for the first issues migration) and do a new `makemigrations` so that we have our "final" database layout as single step migrations.

alanmoo:
Do we then commit that migration set?

PR #35, opened 2017-01-18
Pagination

### LINKED ISSUE #21: add pagination to entries JSON response
We're currently missing a {page,per_page} pair for the `/entries` route, which means there is no way to only grab 10 (or 5 or 20) entries at a time. We'll need to add something like https://docs.djangoproject.com/en/1.10/topics/pagination/#using-paginator-in-a-view to make sure we only send managable sets to the client.

alanmoo:
Closes #21

### DIFF
--- a/public/migrate.html
+++ b/public/migrate.html
@@ -32,9 +32,9 @@
     var currentNonce,
         currentCSRF,
         currentUser;
-      
+
     var entries;
-    
+
     //Sample:
         //[{
         //     "title": "test full entry 1",
@@ -63,7 +63,7 @@
         //     "featured": true,
         //     "issues": ["Decentralization"],
         //     "creators": ["Pomax", "Alan"]
-        // }];
+        // }]
 
     // what to do when "submit" is clicked
     var submitForm = () => {
@@ -118,7 +118,7 @@
       if(entries.length){
         entry = entries.pop()
       } else {finaliseSubmit()}
-      
+
       try {
         data = JSON.parse(data);
 
@@ -137,7 +137,7 @@
       submit.onload = finaliseSumbit
       submit.withCredentials = true;
       submit.send(JSON.stringify(entry));
-    } 
+    }
 
   </script>
   </body>
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -47,6 +47,13 @@ def post_validate(request):
 
     return True
 
+class EntriesPagination(PageNumberPagination):
+    """
+    Add support for pagination and custom page size
+    """
+    page_size = 50
+    page_size_query_param = 'page_size'
+    max_page_size = 1000
 
 class EntryCustomFilter(filters.FilterSet):
     """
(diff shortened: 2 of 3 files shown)


public/migrate.html
```diff
@@ -63,7 +63,7 @@
         //     "featured": true,
         //     "issues": ["Decentralization"],
         //     "creators": ["Pomax", "Alan"]
-        // }];
+        // }]
```
Pomax: =D

pulseapi/entries/views.py
```diff
@@ -47,6 +47,13 @@ def post_validate(request):
 
     return True
 
+class EntriesPagination(PageNumberPagination):
+    """
+    Add support for pagination and custom page size
+    """
+    page_size = 50
```
Pomax: is this default guided by anything?

pulseapi/entries/views.py
```diff
@@ -89,14 +96,16 @@ class EntriesListView(ListCreateAPIView):
 
     **Route** - `/entries`
 
-    **Query Parameters** -
+    #Query Parameters -
 
     - `?search=` - Allows search terms
     - `?tag=` - Allows filtering entries by a specific tag
     - `?issue=` - Allows filtering entries by a specific issue
+    - `?page=` - Page number, default size of 50
```
Pomax: the 'default...' information seems to be on the wrong field here. `Page number, limits returned results to [page*page_size, (page+1)*page_size)`?

pulseapi/entries/views.py
alanmoo: Not really. Suggestion by @mmmavis. Actually I might make it 48 since that's divisible by 2 and 3 nicely, for even load more layouts

pulseapi/entries/views.py
```diff
@@ -47,6 +47,13 @@ def post_validate(request):
 
     return True
 
+class EntriesPagination(PageNumberPagination):
+    """
+    Add support for pagination and custom page size
+    """
+    page_size = 48
+    page_size_query_param = 'page_size'
```
Pomax: Interesting, does the query arg not default to `page_size` without this?

pulseapi/entries/views.py
```diff
@@ -47,6 +47,13 @@ def post_validate(request):
 
     return True
 
+class EntriesPagination(PageNumberPagination):
+    """
+    Add support for pagination and custom page size
+    """
+    page_size = 48
```
Pomax: We might want to file an issue to figure out what a manageable value is based on how many entries the client displays at a time?

pulseapi/entries/views.py
alanmoo: Oh good point- I was going to change it originally but then decided to leave it. Removing.

Pomax: does `page_size` not work as argument unless you specify this?

alanmoo:
Some tests too?

pulseapi/entries/views.py
alanmoo: Turns out if you don't include this, you can't control it. http://www.django-rest-framework.org/api-guide/pagination/#configuration

alanmoo: Turns out if you don't include this, you can't control it. http://www.django-rest-framework.org/api-guide/pagination/#configuration

Pomax:
yes, sorry, some tests that explicitly call the filter route (using some `self.client.get('/entries/bookmarks?page=1&page_size=1')` etc)

alanmoo:
@mmmavis Does the page size default of 48 entries work for you?

pulseapi/entries/views.py
Pomax: haha, that's nuts. Alright then =)

PR #38, opened 2017-01-19
User favorites route

alanmoo:


pulseapi/entries/views.py
```diff
@@ -117,6 +117,17 @@ class EntryView(RetrieveAPIView):
     pagination_class = None
 
 
+class FavoriteEntries(ListAPIView):
```
alanmoo: I didn't inherit from `EntriesListView` because I didn't want this to be POSTable, but we could use some of the filtering options that view has. Thoughts on this?

pulseapi/entries/views.py
```diff
@@ -117,6 +117,17 @@ class EntryView(RetrieveAPIView):
...
+
+    def get_queryset(self):
+        entries = set()
+        user = self.request.user.id
+        for f in UserFavorites.objects.filter(user=user).select_related('entry'):
```
Pomax: small nit: the interpreter doesn't care, but maintainers do, so avoiding single letter variables is generally good practice. Can we call this what it is? (like "entry" or "favorite", although with the user_favs refactor, "bookmark")

pulseapi/entries/views.py
Pomax: Hmm, could we implement this as, itself, a filter? I.e. when users call `/entries/favorites[?args]` Django internally serves up the result for `/entries/[?args]&is_favorited=True`?

alanmoo: Not easily, as per our chat.

alanmoo:
All this needs is tests then?

alanmoo:
I've added some tests. One more look at this?

pulseapi/entries/test_views.py
```diff
@@ -193,3 +193,16 @@ def test_put_bookmark_entry_with_login(self):
...
+        Verify that authenticated users can see a list of bookmarks.
+        """
+        postresponse = self.client.put('/entries/1/bookmark')
+
+        # verify bookmark count is now one
+        bookmarkResponse = self.client.get('/entries/bookmarks/')
+        self.assertEqual(bookmarkResponse.status_code, 200)
+
+        bookmarkJson = json.loads(str(bookmarkResponse.content, 'utf-8'))
+        self.assertEqual(len(bookmarkJson), 1)
```
Pomax: missing newline because github is hilarious that way

Issue #44, opened 2017-01-20
Separate overloaded /nonce into /userstatus and /nonce


Pomax:
In order to make the client work "cleaner", we should split up the `/nonce` route:

- `/nonce` should yield the values required for form posting: `{ nonce: "...", csrftoken: "..." }` or 403 if user is not authed
- `/userstatus` should yield `{ username: "...", email: "...", loggedin: true }` or `{ loggedin: false }` if user is not authed

thoughts @mmmavis @alanmoo ?

alanmoo:
Makes sense to me

PR #47, opened 2017-01-20
Split /nonce into two routes

Pomax:
fixes https://github.com/mozilla/network-pulse-api/issues/44


pulseapi/users/views.py
```diff
@@ -63,16 +63,41 @@ def new_nonce_value(request):
...
+    """
+    name = False
+    email = False
+    loggedin = request.user.is_authenticated()
+
+    if loggedin:
+        name = request.user.name
+        email = request.user.email
+
+    return render(request, 'users/userstatus.json', {
```
alanmoo: Django needs a template to render JSON, it can't just render the JSON straight up?

Pomax: Technically? no. However, using a template means access to default context variables like csrf tokens, so as `nonce` generates off of a template, keeping `userstatus` on a template keeps things consistent between the two routes.

Issue #43, opened 2017-01-20
Polishing migration data


alanmoo:
It looks like 3 entries on the Google Sheet don't have a URL. @xmatthewx can you add something to them, since that's a required field for the new database?

Also, unless I'm missing something, we're dropping a few of the fields from the current entries:

* Favorites (to be regenerated in the new database)
* Status (Will be brought back in the future for community submissions)
* Suggestions & Feedback (does this go to "Internal notes"?)
* Network Connection
* Origin
* Program
* Type
* Backup thumbnail URL (I think this is from moving to hosted images, and fine to drop?)

Is this all correct?

mmmavis:
I'm pretty sure we can drop "Backup thumbnail URL"... as that was one column I was supposed to remove but forgot 😂  Not 100% sure about the other ones but I assume those are okay to drop as well. I'd wait for @xmatthewx 's response.

xmatthewx:
All looks good @alanmoo 

 - **Suggestions**: you can drop this. These were suggestions/feedback about the platform itself. 
 - **Network Connection**: let's map this to `Internal Notes`. 
 - **Origin**. Can we join this and merge this into `Internal Notes`? If not, drop it. 
 -  You can drop the rest.

PR #48, opened 2017-01-20
Migration script

alanmoo:
Running `migrate.py` will output JSON that can be brought in using `migrate.html`. I might not actually merge the sample JSON in, but wanted to give you something to work with for review. Still need some information in #43


Pomax:
this looks like it needs a rebase.

alanmoo:
This should be ready for a proper review. Closes #43

pulseapi/entries/models.py
```diff
@@ -50,8 +51,8 @@ class Entry(models.Model):
         EmailUser,
         related_name='entries',
     )
-    timestamp = models.DateTimeField(
-        auto_now=True,
+    created = models.DateTimeField(
+        default = timezone.now,
```
Pomax: this one is triggering my "timezones are dangerous" senses - we have no guarantee on which timezone we're deployed on, so we may need to make sure this is always, say, UTC, rather than whatever the local time is in the timezone our slug happens to run that hour.

migrate.py
```diff
@@ -0,0 +1,49 @@
+import json
+
+json_data=open('./migrationData.json').read()
```
Pomax: `json_data=open('./migrationData.json', encoding='utf8').read()` (otherwise this will try to read the data as whatever the local terminal's codepage/charmapping is set to. In windows, that means it tries to read it as ANSI)

migrate.py
```diff
@@ -0,0 +1,49 @@
...
+    newEntry["issues"] = arrayify("Issues")
+    newEntry["thumbnail_url"] = entry["Thumbnail URL"]
+    newEntry["internal_notes"] = stringify("Network connection") + " Origin: " + stringify("Origin")
+
+    massagedData.append(newEntry)
+
+with open('massagedData.json', 'w') as outfile:
+    json.dump(massagedData, outfile)
+
+# print(massagedData)
```
Pomax: do we want to keep this in?

PR #54, opened 2017-01-23
add queryset filtering for ids

Pomax:
This adds id filtering so you can call `/entries/ids=1,5,10,12,26` and get the list of just those five entries back.

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -156,7 +156,6 @@ class EntriesListView(ListCreateAPIView):
     - `?page=` - Page number, defaults to 1
     - `?page_size=` - Number of results on a page. Defaults to 48
     """
-    queryset = Entry.objects.public()
     pagination_class = EntriesPagination
     filter_backends = (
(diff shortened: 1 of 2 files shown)


pulseapi/entries/views.py
```diff
@@ -156,7 +156,6 @@ class EntriesListView(ListCreateAPIView):
     - `?page=` - Page number, defaults to 1
     - `?page_size=` - Number of results on a page. Defaults to 48
```
alanmoo: Add the id filter here too

pulseapi/entries/views.py
```diff
@@ -169,6 +168,18 @@ class EntriesListView(ListCreateAPIView):
     )
     serializer_class = EntrySerializer
 
+    # Custom queryset handling: if the route was called as
+    # /entries/ids=1,2,3,4,... only return those entires.
```
alanmoo: `/entries/?ids=1,2,3,4`

PR #55, opened 2017-01-24
Letters aren't all ASCII

alanmoo:
So we should handle that


alanmoo:
I'm not sure if this is necessary but I don't want to remove it because that means testing if everything broke again. Thoughts @Pomax?

Pomax:
it turned out heroku was running python 2.old rather than 3.new - instructing it to always use python 3 solved this issue.

PR #58, opened 2017-01-24
Search query gets tags and creators too

### LINKED ISSUE #57: Search by more fields
Tags and creators, as per https://github.com/mozilla/network-pulse/issues/328#issuecomment-274881754

alanmoo:
Closes #57

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -149,7 +149,7 @@ class EntriesListView(ListCreateAPIView):
 
     #Query Parameters -
 
-    - `?search=` - Allows search terms
+    - `?search=` - Search by title, description, creator, and tag.
     - `?ids=` - Filter only for entries with specific ids. Argument must be a comma-separated list of integer ids.
     - `?tag=` - Allows filtering entries by a specific tag
     - `?issue=` - Allows filtering entries by a specific issue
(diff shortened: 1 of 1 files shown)


pulseapi/entries/views.py
```diff
@@ -168,6 +168,8 @@ class EntriesListView(ListCreateAPIView):
     search_fields = (
         'title',
         'description',
+        'tags__name',
+        'creators__name',
```
Pomax: we probably need to update https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/views.py#L152 so that there's clear documentation on how to actually do searching for these four properties?

pulseapi/entries/views.py
```diff
@@ -149,7 +149,7 @@ class EntriesListView(ListCreateAPIView):
 
     #Query Parameters -
 
-    - `?search=` - Allows search terms
+    - `?search=` - Search by title, description, creator, or tag.
```
Pomax: Based on mattermost chat, this should be "Search based on title, description, creators, *and* tags" rather than *or*, since users don't have a choice here: all fields will be checked.

Based on that description though, if we're searching in creators, are we missing publisher in this list? They're effectively the same category of data.

alanmoo: Updated to use "and".

I don't think publishers are useful, at least at this point in time, especially since it's just staff that can publish at the moment.

Agreed, it may not be ideal, but it's fine for MVP.

PR #64, opened 2017-02-07
Production prep

### LINKED ISSUE #63: turn off detailed error pages on production and staging
I'm not entirely sure how to set it, without digging through the docs, but we need to turn off debug error pages on the production and staging apps on heroku.

alanmoo:
Aims to close #63 and then some


pulseapi/settings.py
```diff
@@ -13,6 +13,19 @@
 import os
 import dj_database_url
 
+import environ
+
+app = environ.Path(__file__) - 1
+root = app - 1
+env = environ.Env(
```
Pomax: curious: why not just the explicit one-line-per-var with an os.environ.get("name", default)?

alanmoo: Because I just copied this from the science api and have no idea what I'm actually doing yet.

Pomax: =D

pulseapi/settings.py
```diff
@@ -16,20 +16,27 @@
...
-DEBUG = True
+DEBUG = env('DEBUG')
 
 # This needs to be a real domain for google auth purposes. As such,
 # you may need to add a "127.0.0.1    test.example.com" to your
 # host file, so that google's redirect works. This is the same
 # domain you will be specifying in your Flow credentials, and
 # associated client_secrets.json
-ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', 'test.example.com,localhost,network-pulse-api-staging.herokuapp.com').split(',')
+ALLOWED_HOSTS = os.getenv('ALLOWED_HOSTS', 'test.example.com,localhost,network-pulse-api-staging.herokuapp.com, network-pulse-api-production.herokuapp.com').split(',')
```
Pomax: you have a rogue space here.

pulseapi/settings.py
```diff
@@ -157,3 +165,12 @@
 CORS_ORIGIN_WHITELIST = os.getenv('CORS_ORIGIN_WHITELIST', 'localhost:3000,localhost:8000,localhost:8080,test.example.com:8000,test.example.com:8080,pulse-react.herokuapp.com').split(',')
 
 CORS_ORIGIN_REGEX_WHITELIST = []
+CSRF_COOKIE_SECURE = True
+CSRF_COOKIE_HTTPONLY = True
+SECURE_BROWSER_XSS_FILTER = True
+SESSION_COOKIE_SECURE = True
+X_FRAME_OPTIONS = "DENY"
```
Pomax: I would love to know why github colours this black, not blue like the rest of the variables...

pulseapi/settings.py
```diff
@@ -157,3 +165,12 @@
...
+SECURE_CONTENT_TYPE_NOSNIFF = True
+SECURE_SSL_REDIRECT = env('SECURE_SSL_REDIRECT')
+SECURE_HSTS_SECONDS = 63072000
```
Pomax: a small nit, but I like being explicit when I'm setting timeouts in seconds. `60 * 60 * 24 * 365 * 2` is a one-time evaluation that makes it much more obvious this code is setting a timeout of two years, measured in seconds.

Alternatively:
```
SECURE_HSTS_SECONDS = datetime.timedelta(weeks=104).total_seconds()
```

pulseapi/settings.py
alanmoo: Bizarre. Single-character-before-an-underscore issue?

pulseapi/settings.py
alanmoo: Good point- I actually pasted the wrong number in here and it should be ~ 6 months anyway.

PR #65, opened 2017-02-08
Adding a thing cade told me to add to make it work

alanmoo:
Because it's caught in a 301 loop on Heroku


alanmoo:
*backs away slowly and tries again*

PR #70, opened 2017-02-09
Env var and login polishing

alanmoo:


templates/users/index.html
```diff
@@ -1,5 +1,10 @@
 {% if user.is_authenticated %}
     <p>Hello {{user.email}}! Would you like to <a href="logout">log out</a>?</p>
 {% else %}
-    <p>Not logged in. Want to <a href="/login?original_url=http://test.example.com:8000">log in</a>?</p>
+    <p>Not logged in. Want to <a id="loginLink">log in</a>?</p>
+    <script>
+    var domain = window.location.origin;
+    var link = document.getElementById('loginLink')
+    link.setAttribute('href', `/login?original_url=${domain}`);
```
Pomax: why just domain? The original_url is "the URL we are on right now", so why not just say `${ window.location.toString() }`?

pulseapi/settings.py
```diff
@@ -165,15 +165,15 @@
...
 
 CORS_ORIGIN_REGEX_WHITELIST = []
-CSRF_COOKIE_SECURE = env('SSL_PROTECTION')
-CSRF_COOKIE_HTTPONLY = env('SSL_PROTECTION')
-SECURE_BROWSER_XSS_FILTER = env('SSL_PROTECTION')
-SESSION_COOKIE_SECURE = env('SSL_PROTECTION')
-X_FRAME_OPTIONS = "DENY"
-SECURE_CONTENT_TYPE_NOSNIFF = env('SSL_PROTECTION')
-SECURE_SSL_REDIRECT = env('SSL_PROTECTION')
+CSRF_COOKIE_HTTPONLY = env('CSRF_COOKIE_HTTPONLY', default=env('SSL_PROTECTION'))
```
Pomax: let's cache this?
```
SSL_PROTECTION = env('SSL_PROTECTION', default=False)
CSRF_COOKIE_HTTPONLY = env('CSRF_COOKIE_HTTPONLY', default=SSL_PROTECTION )
...
```

Issue #80, opened 2017-02-20
A few fields should not be required when posting new entries


mmmavis:
When posting an entry to Pulse API, by design  `csrfmiddlewaretoken`, `nonce`, `title`, and `content_url` should be the only 4 required fields. 

However, currently it seems like Pulse API sees

- `featured`
- `tags`
- `issues`
- `creators`

as required fields as well.  Posting without these 4 fields in the object I'm sending will result status `400` back.

mmmavis:
Also, there's one property `user` that I'm not too sure what it is referring to in https://github.com/mozilla/network-pulse-api/blob/master/public/migrate.html#L125 .

Pomax:
I think the `user` part there isn't actually being used...

Pomax:
Hm, https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/models.py#L40-L55 suggests tags, issues and creators can be empty....

Pomax:
This works fine localhost, but not on the staging herokuapp. Investigation reveals that the staging app is in fact not calling `/nonce` so it's not surprising it's throwing 403, but we should verify this. I've filed https://github.com/mozilla/network-pulse/issues/386 on the client repo

mmmavis:
from http://localhost:8080/migrate.html

<img width="863" alt="screen shot 2017-02-24 at 12 28 20 pm" src="https://cloud.githubusercontent.com/assets/2896608/23319800/ef6ee682-fa8c-11e6-8e30-4177ac7abee3.png">

All of them expect `array` as value. Maybe that's why?

Pomax:
I can't reproduce this, which version of the api are you on and did you remember to migrate your database? (or remove it entirely and rebuilt it with a migrate command)

mmmavis:
I did `git log` and the latest one shows

```bash
commit 5727053cd9d0a60c903373de210c794da8142b00
Author: Alan Mooiman <amooiman@gmail.com>
Date:   Wed Feb 22 15:11:22 2017 -0500

    CSRF_TRUSTED_ORIGINS env var (#88)
```

When I ran `python manage.py runserver` it didn't prompt me to migrate (I did run it a few days ago) . Should I nuke the db and test?

mmmavis:
I did `git log` and the latest one shows

```bash
commit 5727053cd9d0a60c903373de210c794da8142b00
Author: Alan Mooiman <*****@****.com>
[...]

PR #84, opened 2017-02-20
better explanation of the POST entries/ behaviour

Pomax:
addresses https://github.com/mozilla/network-pulse-api/issues/80

documents what payload is needed for the POST /entries operation, including which fields are mandatory and which fields are optional. Also makes it explicit that the `X-CSRFToken` header **must** be used.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ POSTing of entries requires sending the following payload object:
   csrfmiddlewaretoken: required csrf token string obtained from [GET /nonce]
   nonce: required nonce string obtained from [GET /nonce]
 
-  title: required string
+  title: required string (max lengt 140 characters)
   content_url: required url string
 
   description: optional string (max length 600 characters)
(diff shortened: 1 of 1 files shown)


mmmavis:
PR looks awesome ✨

README.md
```diff
@@ -53,16 +53,40 @@ POSTing of entries requires sending the following payload object:
...
 {
-  csrfmiddlewaretoken: '<csrf token>'
-  nonce: '<nonce value obtained from [GET /nonce]>',
-  data: {
-    title: ...
-    ...
+  csrfmiddlewaretoken: required csrf token string obtained from [GET /nonce]
+  nonce: required nonce string obtained from [GET /nonce]
+
+  title: required string
```
mmmavis: oops sorry. Just caught this - we should mention that `title`'s max length is 140 characters

PR #91, opened 2017-02-24
made tags, issues and creators not-required

Pomax:
Fixes https://github.com/mozilla/network-pulse-api/issues/80 by marking the tags, issues and creators as `require=False` in the serializer for Entries

### DIFF
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -41,26 +41,26 @@ class Entry(models.Model):
     tags = models.ManyToManyField(
         Tag,
         related_name='entries',
-        blank=True,
+        blank=True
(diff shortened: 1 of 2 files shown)


pulseapi/entries/models.py
```diff
@@ -51,7 +51,7 @@ class Entry(models.Model):
     creators = models.ManyToManyField(
         Creator,
         related_name='entries',
-        blank=True
+        blank=True,
```
mmmavis: does trailing comma matter?

Pomax: normally yes, in this case no, but I'll remove it anyway

Issue #98, opened 2017-03-10
add image support, aligned with network-api


Pomax:
implementation ticket filed associated with https://github.com/mozilla/network-pulse/issues/363

Based on the work by @cadecairos and @gideonthomas in https://github.com/mozilla/network-api/commit/d692ad329ab95d36028d3d7ebb122fbe9505d83b let's use their solution for image hosting in entries  for the network-pulse-api. As long as we don't insist on users being able to submit images-or-image-urls, but only images, this should be doable reasonably fast.

Pomax:
Some questions:
- a PR is open at the moment that adds a `thumbnail` image field, but leaves the `thumbnail URL` field - both yield URLs when queries through the REST API, but having both seems better than having only one, in case we have content for which there is already a perfect image on the internet.
- The client will need a UI update to allow image data to be passed along with an entry. We're using the same code on the API side as the `network-api` project, so we should be able to look at the client code in `network` to see an example of the data format required for this field. It also feels like it makes sense to have the client present the thumbnail field in a way that lets the user pick one or the other, but not leave both empty, or both used.
- We'll need to hook up the correct S3 bucket values for our staging/production instance
- I noticed that removing images associated with entries does *not* delete the image. As such, we might end up with orphaned data in our bucket and I don't know of a way to make sure we clean up after ourselves (especially for images that should never accidentally make it onto the internet). @cadecairos, @gideonthomas did you look into this?

/cc @wex @xmatthewx

xmatthewx:
Thanks for the update. Much of this sounds good.

> have the client present the thumbnail field in a way that lets the user pick one or the other, but not leave both empty, or both used

We _could_ leave the legacy field in the backend for now if migrating images or updating their URLs is tricky. I thought @alanmoo put these on S3 already. Could we migrate? If not, can we label the input `legacy plz dont use this`

I think doing both introduces complexity for us now and encourages hotlinking that will lead to broken images later. Let's just give users one way to do a thing. It'll be the expected experience.

Pomax:
We can absolutely do that - we can keep it blank for new entries and not "human editable", and then we'll probably have to write a small migration script that ensures that the thumbnail_url data for current database entries are copied over to the new image field (which if I understand correctly, is encoded as a url in the database, too, so that might even be a single copy call)

xmatthewx:
Closing. Tracking the frontend in #424 and the API https://github.com/mozilla/network-pulse-api/issues/98

[...]

PR #99, opened 2017-03-10
add image upload support to to network-pulse-api (nondestructive)

Pomax:
This PR adds a `thumbnail` field to the Entries model, allowing for image data that is either saved locally or to S3, yielding a URL when entries JSON is requested through the REST API.

There are some questions associated with this PR that should be addressed in https://github.com/mozilla/network-pulse-api/issues/98


pulseapi/settings.py
```diff
@@ -12,22 +12,20 @@
...
 
-import environ
+BASE_DIR = root()
+APP_DIR = app()
 
 environ.Env.read_env(os.path.join(BASE_DIR,'.env'))
 env = environ.Env(
     DEBUG=(bool, False),
     SSL_PROTECTION=(bool, False),
+    USE_S3=(bool, True),
```
alanmoo: If we're leaving SSL Protection off by default foe dev purposes, shouln't we do the same for S3?

pulseapi/settings.py
```diff
@@ -53,6 +51,7 @@
     'django.contrib.messages',
     'django.contrib.staticfiles',
     'corsheaders',
+    'storages',
```
alanmoo: Trying to spin up my instance, this can't be found. `requirements.txt` updates needed?

pulseapi/settings.py
Pomax: oh, hah. I had to spin this PR twice, the first time I had it default to `false` but then git happened and I had to start from scratch. Good catch.

pulseapi/settings.py
Pomax: quite, updating.

Pomax:
Due to how the image field works, `thumbnail` and `thumbnail_url` both encode an image URL, but when POSTing data `thumbnail` should be an image, so that's why I've not simply turned `thumbnail_url` into an image field. It seemed functionally equivalent, but naming-wise completely wrong.

alanmoo:
I'm getting an error when trying to hit `/entries`

```
"TemplateSyntaxError at /entries/

'staticfiles' is not a registered tag library."
```

Is this due to `app/staticfiles/` in gitignore or something?

Pomax:
theheq... okay Iooks like removing staticfiles did weird things. I'll have to try this in a clean dir

Pomax:
superceded by #101

PR #101, opened 2017-03-17
Images retry

Pomax:
images should now work (even with `debug-False`). Remember to migrate, then test by running the server through python, and running an http server for the `public` dir. The index.html in that now allows you pick an image rather than a thumbnail url, and posting should show the image in the admin interface.

This PR, if it works, supercedes #99


alanmoo:
The admin interface works but posting images via the separate html page does not save them.

Pomax:
can you explain what "does not save them" means? with `use_s3` set to False, it is supposed to write them to `./media/images/entries/` - when I use the `public/index.html` file to post a test with an image under 2MB, or I use the admin interface to clear the image and then set a new one, files get written correctly there.

Did you remember to `pip install -r requirements` to make sure all the new dependencies get installed?

alanmoo:
The form posting doesn't seem to work in Safari but I'm going to say R+ because that's a front end issue for implementation to deal with. I suspect this has to do with the way the HTML file is submitting the form.

Issue #103, opened 2017-03-20
Unable to post images if POST request content is in json 


mmmavis:
`POST` request from front end has content type `application/json`. I'm getting the following error when posting an entry to Pulse API.

<img width="666" alt="screen shot 2017-03-17 at 3 16 54 pm" src="https://cloud.githubusercontent.com/assets/2896608/24115590/18fcc550-0d61-11e7-815d-99e3de7f45dc.png">

Not sure if there are things we need to add on the back end to enable this or I'm doing something wrong on the front end. I'm opening this ticket so we can get the convo going.

/cc @Pomax

PR #104, opened 2017-03-20
repack thumbnail images so they pass serializer validation

Pomax:
fixes https://github.com/mozilla/network-pulse-api/issues/103

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -1,8 +1,10 @@
 """
 Views to get entries
 """
-
+import base64
 import django_filters
+from django.core.files.base import ContentFile
+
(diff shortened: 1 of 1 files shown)


pulseapi/entries/views.py
```diff
@@ -195,6 +197,27 @@ def post(self, request, *args, **kwargs):
             # invalidate the nonce, so this form cannot be resubmitted with the current id
             request.session['nonce'] = False
 
+            '''
+            If there is a thumbnail, and it was sent as part of an application/json payload,
+            hen we need to unpack a thumbnail object payload and convert it to a Python
```
mmmavis: "then"? :) 🐔

Issue #119, opened 2017-05-25
implement moderation for Entries


Pomax:
Based on the discussions and analysis in https://github.com/mozilla/network-pulse/issues/522, we should implemen Entry moderation functionality using a `ModerationState` model that can be referred to using a `ForeignKey` in the Entry model, allowing us to define a set of moderation states:

- pending
- in review
- approved
- declined

The admin interface should automatically catch on to this and offer pull-down UX for selecting an Entry's moderation state (with `default=pending`), but in order to make sure that the pulse app can be used for moderation we will also need routes that can:

- return the list of moderation states
- set a moderation state for an entry

Pomax:
this has landed, with the routes being added in the pulse-api migration PR

PR #120, opened 2017-05-25
Moderation states

Pomax:
Partially addresses https://github.com/mozilla/network-pulse-api/issues/119 by adding in the `ModerationState` object, which during migration sets up four states (pending, in review, approved, and declined). Entries have been extended to point to the "first" ModerationState (due to the migration, "pending") by default, and the admin view for entries has been extended so that admins can filter on moderation state.

WIP but ready for testing the parts it does already take care of.


pulseapi/entries/migrations/0006_auto_20170525_2235.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
```
Pomax: I don't quite understand where this migration comes from - I pulled master, then checked out a new branch, this migration just gets invented.

pulseapi/entries/migrations/0007_moderationstate.py
```diff
@@ -0,0 +1,35 @@
...
+
+from django.db import migrations, models
+from pulseapi.entries.models import ModerationState
+
+def forwards_func(apps, schema_editor):
+    # We get the model from the versioned app registry;
+    # if we directly import it, it'll be the wrong version
+    # issue = apps.get_model("issues", "Issue")
+    db_alias = schema_editor.connection.alias
+    ModerationState.objects.using(db_alias).bulk_create([
```
Pomax: custom migration seeding, much like issues

pulseapi/entries/migrations/0008_entry_moderation_state.py
```diff
@@ -0,0 +1,36 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.3 on 2017-05-25 22:57
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+from pulseapi.entries.models import ModerationState, Entry, get_default_moderation_state
+
+
+def forwards_func(apps, schema_editor):
```
Pomax: this function basically uplifts all Entries so that if their `is_approved` value is True, their new `moderation_state` value gets updated to point to the "Approved" moderation state

pulseapi/entries/migrations/0009_remove_entry_is_approved.py
```diff
@@ -0,0 +1,19 @@
...
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('entries', '0008_entry_moderation_state'),
+    ]
+
+    operations = [
+        migrations.RemoveField(
```
Pomax: after all the data's been moved over to use moderation_state, the `is_approved` flag is now a soft prop based on the moderation state, and is no longer needed as Boolean field in the model

pulseapi/entries/models.py
```diff
@@ -18,6 +18,25 @@ def entry_thumbnail_path(instance, filename):
...
+        return str(self.name)
+
+
+def get_default_moderation_state():
+    """
+    Helper function to ensure there is a default
+    ModerationState that can be tacked onto Entries. 
+    """
+
+    default_state = ModerationState.objects.all()[0].id
```
Pomax: based on our injection, this is `Pending`. I could make that explicit, too, no strong feelings one way or the other

pulseapi/entries/admin.py
```diff
@@ -30,17 +44,27 @@ class EntryAdmin(admin.ModelAdmin):
...
         'thumbnail_image_tag',
         'bookmark_count',
     )
 
+    ordering = (
+        'created',
+    )
+
+    # this allows us to filter on moderation state in the admin
+    list_filter = (
```
Pomax: super handy. Now admins can view all "pending" entries, for instance.

Pomax:
@alanmoo this switches things over using several migrations just to keep things making sense - I commented various files in the PR to explain what they do (in case it's unclear) but I think this takes care of the moderation state itself. It does not, however, add in the REST routes yet. It should be testable though.

pulseapi/entries/admin.py
```diff
@@ -9,8 +23,8 @@ class EntryAdmin(admin.ModelAdmin):
     """
 
     fields = (
-        'id',
```
Pomax: I should probably put this back in for debug purposes

pulseapi/entries/test_migrations.py
```diff
@@ -0,0 +1,35 @@
+from utility.test_migrations import TestMigrations
+
+class ModerationStateMigrationTestCase(TestMigrations):
+
+    migrate_from = '0006_auto_20170525_2235'
```
Pomax: so.... this seems to fail.

pulseapi/entries/views.py
```diff
@@ -126,11 +128,19 @@ class EntryView(RetrieveAPIView):
     """
     A view to retrieve individual entries
     """
-    queryset = Entry.objects.public()
+
+    #queryset = Entry.objects.public()
+
+    def get_queryset(self):
```
Pomax: we did not have this in place. I don't know what the impact of that was before, but it's kind of important to filter on moz auth or not now!

pulseapi/entries/migrations/0006_auto_20170525_2235.py
alanmoo: I might have changed the help text on `entry` without migrating in a past PR.

pulseapi/entries/views.py
alanmoo: So this now allows the front end to get all entries regardless of moderation state for Mozilla staff?

Pomax: good point. Short answer: yes - if this should be "only approved" then we'll have to update the test for this because staff tests assume "seeing all entries"

pulseapi/entries/models.py
```diff
@@ -98,11 +119,22 @@ def thumbnail_image_tag(self):
...
+
+    # moderation information
+    moderation_state = models.ForeignKey(
+        ModerationState,
+        related_name='entries',
+        default=get_default_moderation_state,
+        on_delete=models.PROTECT
     )
 
+    def set_moderation_state(self, state_name):
```
Pomax: I'm using this primarily to flip entries to "Approved", but this function will also be necessary for when the client has to do moderation calls. The reason it's doing `get_or_create` is mostly because that allows the code to just keep running instead of crashing when it encounters a new state it doesn't know about yet, and it's easy enough to go in and filter for unexpected moderation states in the admin view given that it auto-exposes all possible values as a filter.

pulseapi/entries/migrations/0010_migrate_approval_flags.py
```diff
@@ -0,0 +1,30 @@
+# -*- coding: utf-8 -*-
+# Generated by Django 1.10.3 on 2017-05-25 22:57
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+import django.db.models.deletion
+from pulseapi.entries.models import ModerationState, Entry, get_default_moderation_state
+
+def forwards_func(apps, schema_editor):
+    # We get the model from the versioned app registry;
```
Pomax: clearly this is not the right comment... =D

pulseapi/entries/models.py
```diff
@@ -18,6 +18,25 @@ def entry_thumbnail_path(instance, filename):
...
+        return str(self.name)
+
+
+def get_default_moderation_state():
+    """
+    Helper function to ensure there is a default
+    ModerationState that can be tacked onto Entries.
+    """
+
+    default_state = ModerationState.objects.all()[0].id
```
alanmoo: What's this doing?

alanmoo: Scratch that, I didn't read far enough in another file to understand the migrations.

Pomax:
... what
```
(venv) c:\Users\Mike\Documents\git\collaborations\mozilla\network-pulse-api>manage migrate
could not make use of ModerationState!
CommandError: Conflicting migrations detected; multiple leaf nodes in the migration graph: (0011_remove_entry_is_approved, 0006_auto_20170601_1134 in entries).
To fix them run 'python manage.py makemigrations --merge'
```

alanmoo:
I don't believe I got that error...

Pomax:
for some reason I had a duplicate 0006 but with different timestamps. resolved by killing off the jerk file.

Pomax:
b00g confirmed. Fixed with the most recent commit.

pulseapi/entries/models.py
```diff
@@ -27,7 +46,18 @@ def public(self):
         """
         Return all entries to start
         """
-        return self.filter(is_approved=True)
+        try:
+            # This has to happen in a try/catch, so that migrations
```
alanmoo: Does this mean that once the migration has been run on production, we can remove the try/catch aspect of this?

Pomax: no, it's needed for the migration *path*, so even in dev setting it stays necessary =(

downside of django.

PR #123, opened 2017-06-05
fix migration so it can deal with "duplicates"

### LINKED ISSUE #122: entries migration 0010 assumes individual entries, but may get duplicates
https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/migrations/0010_migrate_approval_flags.py#L14-L17 expects the result to be one, and only one, result. However, as we do not require `{title, content_url}` to be a unique tuple, this migration would crash when it ran into two cases where there were duplicates.

Pomax:
This fixes #122 so that if it finds multiple entries with the same `{title, content_url}` tuple, it won't crash the manage.py process.

Test instructions: see https://github.com/mozilla/network-pulse-api/pull/123#issuecomment-306330071

### DIFF
--- a/pulseapi/entries/migrations/0010_migrate_approval_flags.py
+++ b/pulseapi/entries/migrations/0010_migrate_approval_flags.py
@@ -15,8 +15,9 @@ def forwards_func(apps, schema_editor):
                 title=entry.title,
                 content_url=entry.content_url
             )
-            migrated_entry.moderation_state = approved
-            migrated_entry.save()
+            for m_entry in migrated_entry:
+                m_entry.moderation_state = approved
+                m_entry.save()
 
 
 class Migration(migrations.Migration):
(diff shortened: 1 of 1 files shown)


Pomax:
The result of `get()` is always supposed to be a single result, so this PR switches that over to a `filter()` to get "one or more results", and then iterates over that set explicitly

Pomax:
this comment contains a db.sqlite3 file (zipped) that reflects the database with entries migrated up to and including 0006 but nothing past that, with 4 entries:

```
1: {test, http://example.com, ..., is_approved: 1}
2: {test, http://example.com, ..., is_approved: 1}
3: {utest, http://example.com/utest, ..., is_approved: 1}
4: {utest2, http://example.com/utest2, ..., is_approved: 0}
```
Dropping this file in the network-pulse-api root dir, and then running `python manage.py migrate` will play 0007 through 0011 on top.

[db.sqlite3.gz](https://github.com/mozilla/network-pulse-api/files/1053356/db.sqlite3.gz)

alanmoo:
The attached file appears to have all migrations applied.

image

Pomax:
This one should do the trick [db.sqlite3.gz](https://github.com/mozilla/network-pulse-api/files/1055447/db.sqlite3.gz)

PR #125, opened 2017-06-14
Sync with network api

Pomax:
This copies over the new moderation endpoints and group policy management that network-api uses.

## code changes
- moderation state endpoints
- access control through group policy (three default groups)
- django.sites awareness (necessary for group policies)

## endpoints:

- `GET /api/pulse/entries/moderation-states` returns a JSON object with the information on which moderation states exist by name and id. 
- `GET /api/pulse/entries/?moderationstate=<name of the state>` explicitly gets only the list of entries of a particular moderation state (with the additional query arguments doing additional filtering inside that set before giving you the response). Predefined states are "Pending", "In Review", "Approved" and "Declined".
- `PUT /api/pulse/entries/<entry id>/moderate/<state id>` lets you (attempt to) toggle the moderation state for a specific entry. This requires passing in the state `id`, not the name. Depending on your user rights this can either succeed or return a 403 forbidden. If the entry id does not exist, or the state id does not exist, you will get a 404 instead.

## tests

- edits: updated all tests to work with the new endpoints and code
- additions: three new tests that verify `/moderation-states` works, and that users with "entries.can_change_entry" permissions can moderate, but those without cannot


pulseapi/creators/models.py
```diff
@@ -1,23 +1,28 @@
 """
-The creator field for an entry. Can be empty, just a name, or linked to a pulse user
+The creator field for an entry. Can be empty, just a name,
+or linked to a pulse user
 """
 from django.db import models
+from django.conf import settings
```
alanmoo: Where's this get used?

pulseapi/entries/views.py
```diff
@@ -156,13 +216,18 @@ class EntriesListView(ListCreateAPIView):
...
+                must be a comma-separated list of integer ids.
     - `?tag=` - Allows filtering entries by a specific tag
     - `?issue=` - Allows filtering entries by a specific issue
     - `?featured=True` (or False) - both capitalied. Boolean is set in admin UI
     - `?page=` - Page number, defaults to 1
     - `?page_size=` - Number of results on a page. Defaults to 48
-    - `?ordering=` - Property you'd like to order the results by. Prepend with `-` to reverse. e.g. `?ordering=-title`
+    - `?ordering=` - Property you'd like to order the results by. Prepend with
+                     `-` to reverse. e.g. `?ordering=-title`
+    - `?moderationstate=` - Filter results to only show the indicated moderation
```
alanmoo: Point out here that we need to pass an integer(?)

pulseapi/settings.py
```diff
@@ -48,12 +48,14 @@
...
 
 # Application definition
+SITE_ID = 1
 
 INSTALLED_APPS = [
     'django.contrib.admin',
     'django.contrib.auth',
     'django.contrib.contenttypes',
     'django.contrib.sessions',
+    'django.contrib.sites',
```
alanmoo: What's this do?

pulseapi/entries/views.py
Pomax: it's not, this is a state name

pulseapi/settings.py
Pomax: you don't get group policies without using the sites package

PR #131, opened 2017-06-16
add moderator status to userstatus readout

Pomax:
this adds a "moderator" field to the `userstatus` response, so that clients can tell whether a logged in user has moderation rights or not.

Response will containt `moderator: true` if the user is in the "moderators" group, or if they're a superuser


Pomax:
slips in a small fix for group admin view too, but all it does is remove the calls that also happen in the file that loads in the group admin override, so instead of happening twice, which does nothing because it just unloads and reloads the admin for groups, they're called once.

PR #135, opened 2017-06-19
Add an explicit field for creator ordering

Pomax:
This solves the "how do we track the order of creators" in the most straight-forward way possible: we just literally do that. When the entry is sent over to the API via a POST request, we create a comma-separated list of the names, and store that.

The alternative is something that the Django community agrees is "really really hard" due to the fact that we'd need to track the many-to-many "creator" relation as a relation to "{creator,someuniquvalue}" tuples, such that two entries "{Pomax,...}" are considered *different* creator records, even if they point to the same user. This gets... incredibly complicated.

Especially because all we care about is that we present Pulse users with entries in which the creator list matches the order that whoever posted the entry gave. 

This fixes https://github.com/mozilla/network-pulse/issues/443


alanmoo:
Can you not use https://docs.djangoproject.com/en/1.11/ref/models/options/#order-with-respect-to because it's a ManyToMany relationship?

alanmoo:
If not, https://stackoverflow.com/questions/29036205/order-with-respect-to-for-reverse-relationship looks like it's trying to do something very similar and the answer seems reasonable.

Pomax:
That second thing looks pretty promising - I'd found several SO answers that made this look incredibly complicated. I'll try to redo this based on that SO code

PR #136, opened 2017-06-19
Ordered creator list

Pomax:
Alternative to https://github.com/mozilla/network-pulse-api/pull/135 except that it still doesn't work and touches many files just to get a simple thing done.


pulseapi/entries/admin.py
```diff
@@ -39,7 +39,7 @@ class EntryAdmin(admin.ModelAdmin):
         'featured',
         'internal_notes',
         'issues',
-        'creators',
+        #'creators',
```
Pomax: because `Through` means you can't directly administer

PR #140, opened 2017-06-21
Exposing Issue objects in /issues endpoint

### LINKED ISSUE #139: Change /issues endpoint so it returns an array of Issue objects
As discussed in https://github.com/mozilla/network-pulse/issues/516#issuecomment-302800426, I think it'll be useful if the `/issues` endpoint can return a list of Issue object (`name`, `description`) instead of just an array of issue names.

I got it working locally by trial and error 😂 . But I think there must be something I'm missing... let's discuss it in the PR I'm going to file.

mmmavis:
Fixes #139 

Done by trial and error 😂  but `python manage.py test` said OK. Is there anything I missed? Will this change break any site that are currently using the `/issues` endpoint?

### DIFF
--- a/pulseapi/issues/serializers.py
+++ b/pulseapi/issues/serializers.py
@@ -9,15 +9,12 @@ class IssueSerializer(serializers.ModelSerializer):
     """
     Serializes Issues
     """
-    def to_representation(self, obj):
-        return obj.name
-
     class Meta:
         """
         Meta class. Because it's required by ModelSerializer
         """
         model = Issue
-        fields = ('name', )
+        fields = ('name', 'description')
 
 class IssueDetailSerializer(serializers.ModelSerializer):
     """
--- a/pulseapi/issues/tests.py
+++ b/pulseapi/issues/tests.py
@@ -1,6 +1,6 @@
 import json
 from django.test import TestCase
-
+from pulseapi.issues.models import Issue
 
 class TestIssues(TestCase):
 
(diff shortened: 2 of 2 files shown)


alanmoo:
From a code point, this indeed looks like all that was needed. But since we're switching to an array of objects instead of just an array of strings, the question of "will something break?" is indeed valid- does the pulse front-end access these issues currently? Might need to add in some logic there to temporarily handle the two possible different data types.

As for the network site, I don't believe we pull from the pulse api for issues and didn't find anything on a quick search through the repos. @pomax @gvn can you confirm that the network site isn't hitting `api.mozillapulse.org/issues`?

mmmavis:
@alanmoo oh yea i wasn't sure if Network site or other sites are using the /issues endpoint. Pulse doesn't currently use it so it's fine on my end.

gvn:
Network is currently only using the `/entries` route.

Pomax:
If we're exposing more data, we need an extra test now that calls the issues route, and then checks to make sure that:

1. there are description properties in the response, and
2. descriptions match what they should be.

So in https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/issues/tests.py#L7-L11 we want additional code that does something like:

```
issues = Issue.objects.all()
issue_from_api = issuesJson[0]
issue_from_db = Issue.objects.get(name=issue_from_api.name)
assertEqual(issue_from_db.description, issue_from_api.description)
```

PR #143, opened 2017-06-24
Exposed 'published_by' in /entries

### LINKED ISSUE #138: Expose entry's submitter's name in the /entries endpoint
Context: we want to show the person who submitted the entry on Pulse front end (see this [ticket](https://github.com/mozilla/network-pulse/issues/528))

Currently the `/entries` endpoint doesn't expose the "published by" meta. I realized that "published by" currently is set to the submitter's email and not their full name. I think it's best that we do not expose people's emails so exposing the "published by" meta straight is probably not a good idea. @alanmoo @Pomax can you suggest what we can do? I would love to take this ticket on if it's an easy task. But it seems like there might be some object mapping involved and I can't gauge how difficult it's gonna be for this work.

mmmavis:
Fixes #138

### DIFF
--- a/pulseapi/entries/serializers.py
+++ b/pulseapi/entries/serializers.py
@@ -61,6 +61,8 @@ class EntrySerializer(serializers.ModelSerializer):
                                          queryset=Creator.objects,
                                          required=False)
 
+    # overrides 'published_by' for REST purposes
+    # as we don't want to expose any user's email address
     published_by = serializers.SerializerMethodField()
 
     def get_published_by(self, instance):
--- a/pulseapi/entries/serializers.py
+++ b/pulseapi/entries/serializers.py
@@ -61,6 +61,14 @@ class EntrySerializer(serializers.ModelSerializer):
                                          queryset=Creator.objects,
                                          required=False)
 
+    publisher_name = serializers.SerializerMethodField()
+
+    def get_publisher_name(self, instance):
+        """
+        Get the name of the user who published this entry
+        """
+        return instance.published_by.name
+
     bookmark_count = serializers.SerializerMethodField()
 
     def get_bookmark_count(self, instance):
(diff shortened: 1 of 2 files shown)


mmmavis:
@Pomax This is not a model change nor a view change so I'm not sure where I'm supposed to add test to. 

Also in the `serializer.py`, I see
```py
class Meta:
        """
        Meta class. Because
        """
        model = Entry
        exclude = ('internal_notes', 'published_by',)
```
What is the `Meta` class for? Am I supposed to remove `published_by` from `exclude`?

Pomax:
@mavis the `Meta` class tells django about some of the meta aspects of the object. We definitely want to *keep* `published_by` private because we don't want to expose email addresses. Instead we change the serializer so that it shows a property that doesn't exist as a model property, but will only exist as a REST data property.

As for tests: there is nothing about tests that says they are only for models. Your change has repercussions for what the REST endpoint spits out, so have a look at https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/test_views.py, which has a whole bunch of tests that do exactly that.

In this case you'll want to:
1. get a specific entry in plain Django code (`entry = Entries.objects.last()` for instance)
2. get the REST response for that entry (see https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/test_views.py#L10-L17 for instance)
3. check to see whether `entry.published_by.name` matches the name you find in the response from the REST endpoint

pulseapi/entries/serializers.py
```diff
@@ -61,6 +61,14 @@ class EntrySerializer(serializers.ModelSerializer):
                                          queryset=Creator.objects,
                                          required=False)
 
+    published_by = serializers.SerializerMethodField()
```
mmmavis: is naming this `published_by` okay or we need another name?

mmmavis: @Pomax ^

Pomax: This is fine, although it's a good idea to say here that we're "rebinding" or "overriding" what `published_by` does for REST purposes, compared to what that field does when working directly with a model instance.

Issue #108, opened 2017-03-22
remove `thumbnail_url` from the entries schema


Pomax:
We migrated all `thumbnail_url` data to the new S3-backed `thumbnail` field, and so we can remove the `thumbnail_url` field from the entries schema entirely, with corresponding updates to model.py, admin.py, and serializer.py

This is blocked on https://github.com/mozilla/network-pulse/issues/442, as we do not want to do this until the client has removed all mention and use of the old field name.

Pomax:
the linked issue was resolved, this issue is no longer blocked

PR #145, opened 2017-06-29
remove the deprecated thumbnail_url field

Pomax:
this is a maintenance PR, as the pulse client already uses `thumbnail` rather than `thumbnail_url`.

Closes https://github.com/mozilla/network-pulse-api/issues/108


mmmavis:
Hey @Pomax , is ` pulseapi/entries/migrations/0012_remove_entry_thumbnail_url.py` auto generated? If not could you point me to some docs that explain how to write migration files?

Pomax:
@mavis 100% autogenerated. When you create, edit, or remove models, running `manage.py makemigrations` will make Django create the migration for you automatically in the correct directory.

Pomax:
anything in migrations are to be left alone: migrations run on top of each other, so we added the thumbnail_url at some point in time, and now we're taking it back out. Both those events will be recorded in the migrations path.

Taking it out of the README is a good idea though, no reason to keep it there.

PR #147, opened 2017-07-06
Add local dev step

acabunoc:
Add step: run `pip install -r requirements.txt`


acabunoc:
Let me know if I'm doing this wrong!

Pomax:
nice catch! If it's a docs update then a PR is perfect. (if it's a code update then generally we want an issue filed so we can tie it to a problem description)

PR #150, opened 2017-07-14
Include pagination info in /entries/bookmarks result

### LINKED ISSUE #149: /entries/bookmarks endpoint should return more than just an array
@Pomax @alanmoo 

I just realized `/entries/bookmarks/` endpoint currently only returns an array of entry objects. 

eg., https://network-pulse-api-staging.herokuapp.com/entries/bookmarks/

```js
[
  entryObject1,
  entryObject2,
  ...
]
```

It's missing some essential meta like entries count and pagination info. We should change it so the returned data follows the same schema as what `/entries/` endpoint returns.

e.g., https://network-pulse-api-staging.herokuapp.com/entries/

```js
{
    "count": 169,
    "next": "https://network-pulse-api-staging.herokuapp.com/entries/?page=2",
    "previous": null,
    "results": [
        entryObject1,
        entryObject2,
        ...
    ]
}
```

This is blocking some work on https://github.com/mozilla/network-pulse/issues/310 . 

/cc @hannahkane @xmatthewx

mmmavis:
Fixes #149

### DIFF
--- a/pulseapi/entries/test_views.py
+++ b/pulseapi/entries/test_views.py
@@ -345,8 +345,7 @@ class TestEntryView(PulseStaffTestCase):
 
     def test_bookmarked_entries_view(self):
         """
-        Verify that both authenticated and unauthenticated users can find out
-        what their bookmarked entries are.
+        Verify that authenticated users can see a list of bookmarks.
         """
         # get a legal entry and its associated id
         entries = Entry.objects.all()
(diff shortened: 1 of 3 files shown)


mmmavis:
@Pomax I wonder if we should sort the bookmarked entries from recently bookmarked to least recently bookmarked since that's the order we want to display on the front end. Or maybe we can add a query param `ordering` to allow sorting by different property?

xmatthewx:
Nice work! No need to worry about sorting too much right now. If you want, feel free to file a follow up issue to discuss. Either pub date or fav date is fine for now.

Pomax:
recent-to-last is usually best *unless* you present the user with last-to-recent but start them off on the "last" page with a pagination indicator (because it means that you can bookmark that page-of-bookmarks and it'll always contain the same data). However, I don't think we offer specific page of bookmark linking right now, so sorting most-recent-to-last makes a lot of sense

pulseapi/entries/test_views.py
```diff
@@ -358,8 +358,14 @@ def test_bookmarked_entries_view(self):
...
 
+        # verify data returned has the following properties
         bookmarkJson = json.loads(str(bookmarkResponse.content, 'utf-8'))
-        self.assertEqual(len(bookmarkJson), 1)
+
+        self.assertEqual('count' in bookmarkJson, True)
+        self.assertEqual('previous' in bookmarkJson, True)
+        self.assertEqual('next' in bookmarkJson, True)
+        self.assertEqual('results' in bookmarkJson, True)
+        self.assertEqual(len(bookmarkJson), 4)
```
Pomax: this looks like it's missing the bits to verify that `bookmarkJson.results` still has the correct content

pulseapi/entries/views.py
```diff
@@ -187,11 +186,12 @@ class EntryView(RetrieveAPIView):
 
 
 class BookmarkedEntries(ListAPIView):
+    pagination_class = EntriesPagination
 
     def get_queryset(self):
         user = self.request.user
         bookmarks = UserBookmarks.objects.filter(user=user)
-        return Entry.objects.filter(bookmarked_by__in=bookmarks)
+        return Entry.objects.filter(bookmarked_by__in=bookmarks).order_by('-bookmark_by')
```
Pomax: what is this supposed to order on?

mmmavis: It was done by trial and error so im actually not 100% sure if i did it right. Basically I want the returned entries to be sorted by recently fav’ed to least recently.

I thought `.order_by('-bookmark_by')` is to order the entries by their ids appear in the `users_userbookmarks` table?

mmmavis:
🎉

PR #152, opened 2017-07-14
entries: Add API endpoint to toggle Featured state

arush0311:
Fixes https://github.com/mozilla/network-pulse/issues/572

### DIFF
--- a/pulseapi/entries/urls.py
+++ b/pulseapi/entries/urls.py
@@ -3,6 +3,7 @@ from django.conf.urls import url
 from pulseapi.entries.views import (
     toggle_bookmark,
     toggle_moderation,
+    toggle_featured,
     EntryView,
     BookmarkedEntries,
     ModerationStateView,
@@ -29,6 +30,11 @@ urlpatterns = [
         r'^(?P<entryid>[0-9]+)/moderate/(?P<stateid>[0-9]+)/?',
         toggle_moderation,
         name='moderate'
+    ),
+    url(
+        r'^(?P<entryid>[0-9]+)/feature/?',
+        toggle_featured,
+        name='feature'
     ),
     url(
     	r'^(?P<pk>[0-9]+)/',
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -61,6 +61,31 @@ def toggle_bookmark(request, entryid):
 
 
 @api_view(['PUT'])
+def toggle_featured(request, entryid):
+    """
+    Toggle the featured status of an entry.
+    """
+    user = request.user
+
+    if user.has_perm('entries.can_change_entry'):
+
+        entry = None
+        # find the entry for this id
+        try:
+            entry = Entry.objects.get(id=entryid)
+        except Entry.DoesNotExist:
+            return Response("No such entry", status=status.HTTP_404_NOT_FOUND)
+
+        entry.featured = not entry.featured
+        entry.save()
+        return Response("Toggled featured status.",
+                        status=status.HTTP_204_NO_CONTENT)
+    return Response(
+        "You donot have permission to change entry featured status.",
+        status=status.HTTP_403_FORBIDDEN)
+
+
+@api_view(['PUT'])
 def toggle_moderation(request, entryid, stateid):
     """
     Toggle the moderation state for a specific entry,
--- a/pulseapi/entries/test_views.py
+++ b/pulseapi/entries/test_views.py
@@ -381,6 +381,38 @@ class TestEntryView(PulseStaffTestCase):
         self.assertEqual(len(responseObj), mod_count)
         self.assertEqual(responseObj[0]['name'], "Pending")
 
+    def test_featured_toggle_without_login(self):
+        entry = Entry.objects.all()[0]
+        # ensure that featured status is False
+        entry.featured = False
+        entry.save()
+
+        # ensure that user is logged out
+        self.client.logout()
+
+        url = '/api/pulse/entries/{}/feature'.format(entry.id)
+        response = self.client.put(url)
+
+        self.assertEqual(response.status_code, 403)
+
+        entry.refresh_from_db()
+        self.assertEqual(entry.featured, False)
+
+    def test_featured_toggle_by_staff(self):
+        entry = Entry.objects.all()[0]
+        # ensure that featured status is False
+        entry.featured = False
+        entry.save()
+
+        url = '/api/pulse/entries/{}/feature'.format(entry.id)
+        response = self.client.put(url)
+
+        # did the call succeed?
+        self.assertEqual(response.status_code, 204)
+
+        entry.refresh_from_db()
+        self.assertEqual(entry.featured, True)
+
 
     def test_moderation_toggle_by_staff(self):
         entry = Entry.objects.all()[0]


pulseapi/entries/urls.py
```diff
@@ -31,6 +32,11 @@
         name='moderate'
     ),
     url(
+        r'^(?P<entryid>[0-9]+)/featured/?',
+        toggle_featured,
+        name='featured'
```
Pomax: this `name` is used to resolve urls, so as this is a toggle, keeping it a verb is a good idea. Let's change this to `feature` instead of `featured`

pulseapi/entries/views.py
```diff
@@ -62,6 +63,28 @@ def toggle_bookmark(request, entryid):
 
 
 @api_view(['PUT'])
+@permission_classes([IsAdminUser])
```
Pomax: this is not quite right, because it assumes only admin users may change this.

However, permissions are handled through policy rules (based on Django "groups"), so rather than using an admin decorator, you will need to use a solution similar to https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/entries/views.py#L65-L100 where the user's permissions are checked to see if they have entry-modification rights.

pulseapi/entries/views.py
```diff
@@ -8,13 +8,14 @@
 
 from rest_framework import (filters, status)
 from rest_framework.decorators import (
-    detail_route, api_view
+    detail_route, api_view, permission_classes
```
Pomax: `permission_classes` is now an unused import, let's remove it again.

pulseapi/entries/views.py
```diff
@@ -8,13 +8,14 @@
...
 )
 from rest_framework.generics import (
     ListCreateAPIView, RetrieveAPIView, ListAPIView
 )
 from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
+from rest_framework.permissions import IsAdminUser
```
Pomax: `IsAdminUser` is now an unused import, let's remove it again.

pulseapi/entries/views.py
```diff
@@ -61,6 +62,31 @@ def toggle_bookmark(request, entryid):
 
 
 @api_view(['PUT'])
+def toggle_featured(request, entryid):
```
Pomax: It is now worth doing a little refactoring: we have two functions that are effectively identical except for the model value they update: this is the perfect time to create a single function that does the bulk of the work, and two functions that rely on that but house the "different" bits of code, so that we get something like:

```python
def get_entry_to_modify(request, entryid):
    '''
    This function will return an Entry object if the user
    associated with the request has entry change rights,
    or it will return a (non-200) Response object to be
    sent back as server response when they do not.
    '''
    
    # all the code that checks user's permissions
    # and either fetches an entry or forms a non-2xx
    # Response goes here

@api_view(['PUT'])
def toggle_featured(request, entryid):
    entry = get_entry_to_modify(request, entryid)

    if isInstance(entry, Response) is True:
        return target

    entry.featured = not entry.featured
    entry.save()

    return Response("Toggled featured status.", status=status.HTTP_204_NO_CONTENT)

@api_view(['PUT'])
 def toggle_moderation(request, entryid, stateid):
    entry = get_entry_to_modify(request, entryid)

    if isInstance(entry, Response) is True:
        return target

    entry.moderation_state = ModerationState.objects.get(id=stateid)
    entry.save()

    return Response("Toggled moderation state.", status=status.HTTP_204_NO_CONTENT)
```

Pomax:
In addition to this code, you will also need to add a new test case to `pulseapi/tests.py`. Have a look at how bookmarking is tested in that file: testing feature toggling should basically work in the same way (with separate test cases so that failures clearly indicate which test case they're failing for)

arush0311:
@pomax i did make some changes, but instead of creating my own functions i used django and drf ones

pulseapi/entries/views.py
```diff
@@ -26,77 +29,62 @@
 from pulseapi.utility.userpermissions import is_staff_address
 
 @api_view(['PUT'])
+@permission_classes([IsAuthenticated])
```
Pomax: I'm not a fan of this solution: we are not checking whether a user has a permission, we are checking whether a user is authenticated or not, so let's not do this. Let's keep using `user.is_authenticated()`

pulseapi/entries/views.py
Pomax: This is still an unused import, please remove it.

pulseapi/entries/views.py
```diff
@@ -26,77 +29,62 @@
...
     Toggle whether or not this user "bookmarked" the url-indicated entry.
     This is currently defined outside of the entry class, as functionality
     that is technically independent of entries themselves. We might
     change this in the future.
     """
-    user = request.user
 
-    if user.is_authenticated():
-        entry = None
+    user = request.user
```
Pomax: This issue is not about changing the code for bookmarking. Please don't overload PRs to do "more than they should". If you want to improve the bookmarking code, file a new issue for that and then a separate PR to effect the desired update.

pulseapi/entries/views.py
```diff
@@ -26,77 +29,62 @@
...
+    Toggle the featured status of an entry.
+    """
+    entry = get_object_or_404(Entry, id=entryid)
+    entry.featured = not entry.featured
+    entry.save()
 
-        return Response("Toggled bookmark.", status=status.HTTP_204_NO_CONTENT)
-    return Response("Anonymous bookmarks cannot be saved.", status=status.HTTP_403_FORBIDDEN)
+    return Response("Toggled featured status.",
+                    status=status.HTTP_204_NO_CONTENT)
```
Pomax: I recommended using a function for this response object because it means we no longer have to break code just because it's 4 characters too long. If we're putting in a linebreak here, let's use that function call instead and improve all `return Response` lines.

pulseapi/entries/views.py
```diff
@@ -26,77 +29,62 @@
...
 
 
 @api_view(['PUT'])
+@permission_required('entries.can_change_entry', raise_exception=True)
```
Pomax: Exceptions break the server, so please don't add a `raise_exception=True`. This code was not broken, so let's leave it alone, and file an improvement issue instead so that we can look at refactor as its own job, not slipped into a PR that adds feature toggling.

pulseapi/entries/views.py
arush0311: @pomax i removed all refactor code from this pr..

pulseapi/entries/views.py
arush0311: `permission_classes` have nothing to do with `django permission objects`. it just checks whether a user has *permission* to to access that api point. which in this case is by checking if he is authenticated or not. It was a convenience method designed to replace `user.is_authenticated()` flow

pulseapi/entries/views.py
arush0311: I think its better to break the code into multiple lines than create a new method as more levels increases complexity and reduces readability. And in this case it doesnot provide any more abstraction or dryness

pulseapi/entries/views.py
arush0311: raise_exception will send a 403 response code to client, otherwise the user is redirected to `LOGIN_URL`. In django certain exceptions are linked to certain views, so it wont restart the server..

Pomax:
Thanks for the rewrite - let's look at the refactoring as part of https://github.com/mozilla/network-pulse-api/issues/158

PR #154, opened 2017-07-18
entries: Make `get_involved` and `interest` fields searchable

arush0311:
Fixes https://github.com/mozilla/network-pulse/issues/594

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -217,7 +217,8 @@ class EntriesListView(ListCreateAPIView):
 
     #Query Parameters -
 
-    - `?search=` - Search by title, description, creator, and tag.
+    - `?search=` - Search by title, description, get_involved, interest,
+                   creator, and tag.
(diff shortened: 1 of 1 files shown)


arush0311:
Do we want to make `internal_notes` searchable?

xmatthewx:
> Do we want to make internal_notes searchable?

No. Good question. That is strictly for staff use internally.  Thanks!

pulseapi/entries/views.py
```diff
@@ -241,6 +241,8 @@ class EntriesListView(ListCreateAPIView):
     search_fields = (
         'title',
         'description',
+        'get_involved',
```
Pomax: please also update https://github.com/arush0311/network-pulse-api/blob/79bbbd7b9f80543dbdebfa97738f88ab8eaf17e5/pulseapi/entries/views.py#L220 so that the fact that these are searched is documented when people open the DRF page on `/entries`

Pomax:
thanks!

PR #156, opened 2017-07-20
Bulk bookmarking

### LINKED ISSUE #118: Add route to bookmark a list of entries

mmmavis:
Fixes #118

### DIFF
--- a/pulseapi/entries/test_views.py
+++ b/pulseapi/entries/test_views.py
@@ -405,36 +405,7 @@ class TestEntryView(PulseStaffTestCase):
         Verify that authenticated users can bookmark a list of entries.
         """
 
-        # get a legal entry and its associated id
-        entries = Entry.objects.all()
-        entry = entries[0]
-        id = entry.id
-
-        url = '/api/pulse/entries/bookmarks/?ids=' + str(id)
-        payload = self.generatePostPayload()
-
-        postresponse = self.client.post(url,payload)
-        self.assertEqual(postresponse.status_code, 204)
-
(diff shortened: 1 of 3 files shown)


pulseapi/entries/test_views.py
```diff
@@ -380,6 +380,115 @@ def test_bookmarked_entries_view(self):
...
+        self.client.logout()
+        url = '/api/pulse/entries/bookmarks/?ids=' + str(id)
+        postresponse = self.client.post(url)
+        self.assertEqual(postresponse.status_code, 403)
+
+        # verify bookmark count is zero
+        bookmarks = entry.bookmarked_by.count()
+        self.assertEqual(bookmarks, 0)
+
+    def test_post_bookmark_entries_with_login(self):
```
Pomax: What is being tested in this test? For toggling bookmarks we already have `/entries/:id/bookmark` so if this simply tests that bulk bookmarking but with only a single entry works, then we should combine that with posting a comma delimited list.

pulseapi/entries/views.py
```diff
@@ -197,7 +198,67 @@ def get_queryset(self):
         bookmarks = UserBookmarks.objects.filter(user=user)
         return Entry.objects.filter(bookmarked_by__in=bookmarks).order_by('-bookmarked_by__timestamp')
 
-    serializer_class = EntrySerializer
+    # When people POST to this route, we want to do some
```
Pomax: This is going to need an update to the README.md too, to explain that this route exists, how to call it, and what can be expected back =)

PR #161, opened 2017-07-21
Ignore invalid id instead of returning status 400

### LINKED ISSUE #157: Adjustment to bulk bookmarking response
Currently if the `ids` passed to the `/entries/bookmarks/ids=` route contains an invalid id (non-existent IDs and non-numerical IDs) we send a HTTP 400 Bad Request response back as soon as we spot the invalid entry id. However, some entries might have already been bookmarked and sending a HTTP 400 back can be misleading.

Let's make the API "pass over" non-existent IDs and non-numerical IDs instead of sending a HTTP 400 back.

mmmavis:
Fixes #157

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -340,11 +340,8 @@ class EntriesListView(ListCreateAPIView):
         ids = self.request.query_params.get('ids', None)
 
         if ids is not None:
-            try:
-                ids = [int(x) for x in ids.split(',')]
-                queryset = queryset.filter(pk__in=ids)
-            except:
-                pass
+            ids = [int(x) for x in ids.split(',')]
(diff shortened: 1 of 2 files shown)


pulseapi/entries/views.py
```diff
@@ -214,14 +214,15 @@ def post(self, request, *args, **kwargs):
...
-                    )
+                    return int(id)
+                except ValueError:
+                    return id
+
+            def remove_str(id):
+                return isinstance(id, int)
+
 
             def bookmark_entry(id):
```
Pomax: we should be able to just fix up this function by swapping `except Entry.DoesNotExist:` for just `except:`

Pomax:
nice! this should help with the bulkmarking

PR #162, opened 2017-07-21
Add helptypes in a way similar to issues

Pomax:
addresses the server side of https://github.com/mozilla/network-pulse/issues/601


Pomax:
- README: absolutely, let me update the PR
- naming conventions: the internet tends to favour lowecasewords as well as words-with-hyphens, so that's mostly what we're using. Relying on case sensitivity InThisWay is generally a bad idea since not every OS and httpd even supports case sensitivity, and snake case is really just for programmers... it's a python convention (I only just realised that that snake case makes sense because a python is a kind of snake... wat)  but should not make its way into API urls.

PR #164, opened 2017-07-25
Exposing 'helptypes' of an entry

### LINKED ISSUE #163: 'helptypes' of an entry not being exposed
It seems like `helptypes` (`help_types`?) associated with an entry is not being exposed yet https://network-pulse-api-staging.herokuapp.com/entries/

mmmavis:
Fixes #163


mmmavis:
@Pomax for consistency I think we should lowercase one the the help type options "Take Action"(change it to "Take action"). How do I do that?

Pomax:
@mmmavis we can use CSS `text-transform: capitalize` for that.

mmmavis:
@Pomax oh yet it's just https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/helptypes/migrations/0001_initial.py#L48 seems to be the only one that has 2 caps in the string. I wasn't sure if we would wanna fix it or not.

mmmavis:
🎉

PR #167, opened 2017-07-27
add tag cleanup migration and admin improvement

Pomax:
This adds tag cleanup in the form of:
- an operation that collapses multiple instances of a tag into a single cased tag
- an operation that forces all tags (after deduplication) to lower case
- improvements to the tag admin, by:
  - showing "how many entries use this tag"
  - directly showing all entries when viewing a specific tag's admin page

This can be tested against the following database: 
[test_db.zip](https://github.com/mozilla/network-pulse-api/files/1181294/test_db.zip)

This database contains the tags as used in the production database, with generic entries for each tag, with random additional tag links into each entry.

This patch reduces the production tag count from 534 to 444 tags.


mmmavis:
@Pomax I think it would be helpful if we could chat over this on Vidyo after tomorrow's stand. I'm a bit confused what tags we are gonna keep.

Using entry id 140 and 141 as examples

### Entry no.140

Before migration....
screen shot 2017-07-31 at 5 43 41 pm

After migration....
This entry no longer has "wwl" tag associated with it.
screen shot 2017-07-31 at 5 44 05 pm


### Entry no.141

Before migration....
screen shot 2017-07-31 at 5 47 58 pm


After migration....
This entry no longer has "iot" tag associated with it.
screen shot 2017-07-31 at 5 48 03 pm

Issue #133, opened 2017-06-19
split EmailUser into a user and a userprofile model


Pomax:
This will make it much more easy to merge into another codebase later

- user: only contains username, email, password
- user profile: contains things like faves, bio, etc.

xmatthewx:
<img width="1314" alt="user-profile-model" src="https://user-images.githubusercontent.com/53411/28840434-ebf4ff5a-76c4-11e7-8698-8d86443622b4.png">

My diagram doesn't include Favs. Are there other things missing that should be represented?

Pomax:
PR landed!

PR #171, opened 2017-08-04
Move bookmarks from EmailUser to UserProfile

Pomax:
Addresses https://github.com/mozilla/network-pulse-api/issues/133

This PR sets up a UserProfile model, and moves user `bookmarks` over from the EmailUser model to this UserProfile model instead. This will let us both start padding out the UserProfile, and clean up the EmailUser to the point where we can hopefully replace it with a Django-managed user instead.

How to test this PR:

1. follow the setup instructions in the README for making sure you can use/log in
2. grab this database file: [db.tar.gz](https://github.com/mozilla/network-pulse-api/files/1201641/db.tar.gz) and unpack it to the root. This is a database with migrations applied equal to the pre-this-PR codebase, with a bare minimum of faves stored
3. run `manage migrate`, then `manage test`. The migration will likely ask you to delete the stale database users_userbookmarks: that is fine. Running `manage test` should show 41 tests passing.
4. run `manage runserver` and hit up your http://test.example.com:8000/admin and have a look at the profile/bookmark data, as well as what's in the emailuser entries. It should all look kosher.


pulseapi/profiles/models.py
```diff
@@ -0,0 +1,57 @@
...
+    bookmarks = models.ManyToManyField(
+        'entries.Entry',
+        through='profiles.UserBookmarks'
+    )
+
+    def __str__(self):
+        return 'profile for {}'.format(self.user.email)
+
+
+class UserBookmarks(models.Model):
```
gideonthomas: nit: probably need a `Meta` class to specify the verbose name as `Bookmark` so it doesn't add the extra `s` - `Bookmarkss`

pulseapi/users/admin.py
```diff
@@ -3,33 +3,46 @@
...
+            id=profile.id,
+        )
+
+        return format_html(html)
+
+    def bookmarks(self, instance):
+        """
+        Show all bookmarked entries as a string of titles. In the future we should make them links.
+        """
+        profile = UserProfile.objects.get(user=intance)
```
gideonthomas: typo: `instance`

pulseapi/users/admin.py
```diff
@@ -3,33 +3,46 @@
...
 class EmailUserAdmin(admin.ModelAdmin):
     """
     Show a list of entries a user has submitted in the EmailUser Admin app
     """
-    fields = ('password', 'last_login', 'email', 'name', 'entries','bookmarks', 'is_staff', 'is_superuser')
-    readonly_fields = ('entries','bookmarks')
-
-    # this allows us to create/edit/delete/etc bookmarks:
-    inlines = [ UserBookmarksInline ]
+    fields = ('password', 'last_login', 'email', 'name', 'is_staff', 'is_superuser', 'profile', 'entries','bookmarks', )
```
gideonthomas: so we still show bookmarks in the EmailUser, not in the UserProfile?

pulseapi/profiles/admin.py
```diff
@@ -0,0 +1,19 @@
+from django.contrib import admin
+from .models import UserProfile, UserBookmarks
+
+class UserProfileAdmin(admin.ModelAdmin):
+    """
+    Show a list of entries a user has submitted in the EmailUser Admin app
```
gideonthomas: We'll change the comments in a future pr when we add the actual fields to a user profile?

pulseapi/users/admin.py
Pomax: Both. It's useful for admins to see the bookmark count in the user view (although I took out the inline, there is no reason to see each bookmark as editable entity...)

pulseapi/profiles/admin.py
Pomax: we will, but that doesn't make this the right comment - fixing to say it shows the profile-associated user, if there is one set.

PR #175, opened 2017-08-11
Create Profile model

Pomax:
Implementation of the Profile model as discussed in https://github.com/mozilla/network-pulse/issues/621

This is a work in progress, review comments are fine, but do not merge.


Pomax:
@gideonthomas I've marked you as review mostly so that there's an "is this code sane?" pass before more work is put into this.

pulseapi/profiles/models.py
```diff
@@ -1,19 +1,117 @@
 from django.db import models
+from django.utils.html import format_html
+
+
+def entry_thumbnail_path(instance, filename):
+    return 'images/user-avatars/{timestamp}{ext}'.format(
+        timestamp=str(timezone.now()),
```
gideonthomas: `timezone` needs to be imported?

pulseapi/profiles/models.py
```diff
@@ -1,19 +1,117 @@
...
+    - Some cities lie in to countries. I know, that's crazy, but some do.
+
+    - Some people live or work or associate with more than one location,
+      especially if they live near an easily crossed border such as in
+      many European countries.
+
+    Making this do the right thing is going to take more work than any
+    initial model implementation can achieve.
+    """
+    city = models.CharField(max_length=500, blank=False, null=True)
```
gideonthomas: nit: `blank` is `False` by default

gideonthomas: one thing to note about setting this to `False` is that for existing entries, once you open it with the intent of changing another property, it will fail to save until you also add a value for this field.

pulseapi/profiles/models.py
```diff
@@ -1,19 +1,117 @@
...
+            country=self.country
+        )
+
+
+# We want users to be able to give any number of "social" URLs that are
+# relevant to them, such as facebook, linkedin, twitter, blog url,
+# github profile, etc. etc.
+#
+class SocialUrl(models.Model):
+    url = models.URLField()
```
gideonthomas: `max_length` for this is `200`. You might want to increase that.

pulseapi/profiles/models.py
```diff
@@ -23,9 +121,72 @@ class UserProfile(models.Model):
...
+        max_length=500,
+        blank=True,
+        null=True
+    )
+    custom_name.short_description = 'Custom user name'
+
+    # Accessing the Profile-indicated name needs various stages
+    # of fallback.
+    def name(self):
+        if not self.custom_name:
```
gideonthomas: the problem with using `not` here is that `if not "   "` evaluates to `False` so it will return the whitespaces. Don't know if that is expected behavior or not.

pulseapi/profiles/models.py
Pomax: good point

pulseapi/profiles/models.py
Pomax: good catch! are you flake8-ing?

pulseapi/profiles/models.py
Pomax: that's a good point, I should probably expand this to have a second if for `self.custom_name.strip()`

pulseapi/users/admin.py
```diff
@@ -6,21 +6,32 @@
 from django.utils.html import format_html
 
 from .models import EmailUser
+from pulseapi.entries.models import Entry
 from pulseapi.profiles.models import UserProfile, UserBookmarks
+from pulseapi.utility.get_admin_url import get_admin_url
+
+class EntryInline(admin.TabularInline):
+    model = Entry
+    verbose_name = 'entries'
```
gideonthomas: wouldn't this be `verbose_name_plural`, with `verbose_name` being `entry`?

Pomax: I though that for inlines (which don't exist in the admin nav panel) those were the same, but maybe they're not.

gideonthomas: hmm I'm not sure, I can't find docs suggesting that though.

Pomax: easy enough to change.

pulseapi/profiles/models.py
```diff
@@ -24,6 +90,95 @@ class UserProfile(models.Model):
...
+
+    thumbnail_image_tag.short_description = 'Thumbnail preview'
+
+    # Which issues does this user care about/are they involved in?
+    issues = models.ManyToManyField('issues.Issue')
+
+    # we allow users to indicate several possible predefined service URLs
+    twitter = models.URLField(
+        max_length=2048,
+        blank=True
```
gideonthomas: wouldn't you also need this to be `null=True` if you're allowing blank content but not auto-generating anything.

Pomax: hm, makemigrations did not complain about that, doesn't `blank=true` set up a default value? (null is not allowed, but blank is, so when it creates one it'll just be whatever is an empty url field, presumably a blank string?)

gideonthomas: hmm interesting. So I think for string fields, it does default to empty string so this might not be a problem.

PR #178, opened 2017-08-14
fix tags with commas and trim tags when saving

Pomax:
fixes https://github.com/mozilla/network-pulse/issues/649


Pomax:
yeah we that would be too intricate, since we'd also have to manually hook up entries to the correct post-split tags. This is functionality purely for automation.

PR #187, opened 2017-08-24
Order creators in entries by post date

gideonthomas:
This should force it to be ordered by the order in which the creator-entry relations are added in the through model (since you specify ordering in the model itself) instead of the order in the creator model. From what I read, the ordering of through models is only respected on querysets that are queried on that model and not the creator model. So this patch queries the OrderedCreatorRecord model instead of the Creator model and so the ordering follows whatever you specified on that model.

I also changed the serializer to serialize the relation as a creator (the same way the CreatorSerializer works). Feel free to revert that if not needed.


gideonthomas:
Tests are failing because it needs to run makemigrations (I didn't commit it in case you need to change something)

PR #205, opened 2017-09-12
expose submitter_profile_id for entries

Pomax:
This adds a `submitter_profile_id` property to entries, which can be used to crosslink entries to the posting user's profile in the client

### DIFF
--- a/pulseapi/entries/serializers.py
+++ b/pulseapi/entries/serializers.py
@@ -90,13 +90,17 @@ class EntrySerializer(serializers.ModelSerializer):
         return instance.published_by.name
 
     # "virtual" property so that we can link to the correct profile
-    submitter_id = serializers.SerializerMethodField()
+    submitter_profile_id = serializers.SerializerMethodField()
 
-    def get_submitter_id(self, instance):
+    def get_submitter_profile_id(self, instance):
         """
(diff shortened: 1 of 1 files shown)


pulseapi/entries/serializers.py
```diff
@@ -89,6 +89,19 @@ def get_published_by(self, instance):
...
         return instance.published_by.name
 
+    # "virtual" property so that we can link to the correct profile
+    submitter_profile_id = serializers.SerializerMethodField()
+
+    def get_submitter_profile_id(self, instance):
+        """
+        Get the id for the user who published this entry
+        """
+        profiles = instance.published_by.profile.all()
```
Pomax: this is a little odd, but a consequence of how we ended up implementing profiles. They're technically a many-to-many relation (and I already can't remember why) so accessing `instance.published_by.profile` yields a RelationManager, not an actual UserProfile object.

gideonthomas: it looks like it's a one-to-many relation, i.e. 1 User has many profiles. Yeah weird, it probably should have been 1:1 (optional relationship due to orphaned org profiles).

gideonthomas:
thanks @Pomax!

PR #208, opened 2017-09-15
Allow fields in myprofile payload to be blank

### LINKED ISSUE #206: PUT /myprofile payload question
(Feel free to change the ticket title. I'm having trouble making it more descriptive)

Questions regarding optional fields in `profile` and the PUT /myprofile payload...

I just tried updating a profile with the following payload but got status 400 back.

screen shot 2017-09-14 at 1 17 46 pm
screen shot 2017-09-14 at 1 18 02 pm

Not sure how things work in the backend. It seems a bit odd that I can't include a key with empty value for optional fields. (Use case: someone wants to clear the `twitter` field because he longer uses that Twitter page.) My assumption is Pulse API will take the payload and does per key-pair overwrite instead of "take this payload entirely and overwrite the entire profile with that payload"? Is this true? @Pomax @gideonthomas

gideonthomas:
Fixes #206 

You can now pass empty strings to these fields and they will be considered valid values. Tbh, there's probably a better way to do this with `null` since django treats `null` and `""` as two distinct values for strings while on the front-end we treat them as the same. But I think this works fine for now.

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -26,11 +26,13 @@ class UserProfileSerializer(serializers.ModelSerializer):
     """
     user_bio = serializers.CharField(
         max_length=140,
-        required=False
+        required=False,
+        allow_blank=True,
     )
     custom_name = serializers.CharField(
         max_length=70,
-        required=False
+        required=False,
+        allow_blank=True,
     )
     is_group = serializers.BooleanField()
     issues = serializers.SlugRelatedField(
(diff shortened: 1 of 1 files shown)


pulseapi/profiles/serializers.py
```diff
@@ -26,11 +26,13 @@ class UserProfileSerializer(serializers.ModelSerializer):
...
+        required=False,
+        allow_blank=True,
     )
     custom_name = serializers.CharField(
         max_length=70,
-        required=False
+        required=False,
+        allow_blank=True,
     )
     is_group = serializers.BooleanField()
```
mmmavis: Front end UI doesn't have a field to toggle `is_group` (I think this is an admin only control that's only visible on Django dashboard?). Can serialize do something like "if the value is `null` or the `is_group` key isn't presented, keep the `is_group` value in db unchanged" ?

xmatthewx: > Front end UI doesn't have a field to toggle is_group (I think this is an admin only control that's only visible on Django dashboard?)

Yes. It's just a moderator/admin feature we'll use via Django, for when staff wants to create a profile for something like Ford Foundation. No real functionality now, but useful down the road.

Pomax:
@mmmavis given matthew's response, is this R+ now?

mmmavis:
@Pomax ummm I think we still need to update this PR so that `is_group` is an optional key. Right now the payload must include `is_group` key and the value must be a boolean, otherwise API returns a `400` back.

Pomax:
ahh, gotcha. @gideonthomas that should be a pretty trivial update to the model, right?

gideonthomas:
yeah I can make this change

PR #209, opened 2017-09-19
adding npm scripts for better dev ergonomics

gvn:
🎉 🎉 🎉

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,0 +1,5 @@
+{
+  "name": "network-pulse-api",
+  "version": "1.0.0",
+  "lockfileVersion": 1
+}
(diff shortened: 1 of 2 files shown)


gvn:
It shouldn't cause any problems...

gideonthomas:
should be fine

Pomax:
One bit problem is that this is unix-specific. We might need to actually write a small script that does a `process.platform` and calls either `source ven/bin/activate` or `venv\Scripts\Activate` depending on the OS found.

gvn:
Let's do that as a follow on since this is just an extra shortcut right now.

Pomax:
but... I'm a windows dev =(

I mean I'm fine with follow-up of course, but it's always that little bit disheartening to see windows getting relegated to an also-ran to be done later. Filed https://github.com/mozilla/network-pulse-api/issues/210 for the follow-up

PR #212, opened 2017-09-22
Send profile info for creators when fetching entries

gideonthomas:
Addresses part of mozilla/network-pulse#624

This is most of the work we need to do for the front-end to distinguish between creators with profiles and creators without profiles.

Things that still need to be done:
- Fixing how we get a users own profile, since they can own multiple profiles. It's kinda hacky to rely on `is_group` since that would mean that admins who manage profiles have to manually make sure that for a user that has multiple profiles, only one of those profiles has `is_group` set to `False` viz. their own profile
- Exposing a list of creators so that the front-end can use it for auto-complete
   - Part of this is also fixing how creators are created when posting a new entry - not just doing it based on name but also based on profile (automatically linking a profile to a creator)

Will file issues for the above!

### DIFF
--- a/pulseapi/entries/serializers.py
+++ b/pulseapi/entries/serializers.py
@@ -82,7 +82,7 @@ class EntrySerializer(serializers.ModelSerializer):
 
     # Although this field has similar results to the field above (it's just
     # serialized differently), we create a new field vs. overriding the field
-    # above so that we maintain backwards compatibility
+    # above so that we maintain backward compatibility
     creators_with_profiles = serializers.SerializerMethodField()
 
     def get_creators_with_profiles(self, instance):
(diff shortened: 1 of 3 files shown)


pulseapi/entries/serializers.py
```diff
@@ -79,6 +80,27 @@ def get_creators(self, instance):
         """
         return [ocr.creator.name for ocr in instance.related_creators.all()]
 
+    # Although this field has similar results to the field above (it's just
+    # serialized differently), we create a new field vs. overriding the field
+    # above so that we maintain backwards compatibility
```
Pomax: nit: 'backward'

pulseapi/users/models.py
```diff
@@ -77,6 +77,14 @@ def get_short_name(self):
     def clean(self):
         pass
 
+    def get_self_profile(self):
+        """
+        Returns the profile belonging to this user object since
+        a user can have multiple profiles attached to it that
+        represent organizations
+        """
+        return self.profile.filter(is_group=False).last()
```
Pomax: hm, what are we relying on in this code? "there is only one profile that is not a group profile", and `last()` is just a convenience accessor stand-in for `[0]`?

gideonthomas: yeah pretty much, it's a hack until we fix point 1 from my comment above. Will file to figure out how to handle that

PR #216, opened 2017-09-27
Test creators being serialized correctly for entries

gideonthomas:
This is the last bit for https://github.com/mozilla/network-pulse/issues/624.

We only need to test the serializer (vs. the whole route since other tests already test this) since that is where we have our custom logic.

I had to actually create creators and link them to entries in the setup (which broke a few tests re: creators, so I fixed them).

### DIFF
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -10,17 +10,6 @@ class TestCreators(PulseStaffTestCase):
 
     def test_creator_filtering(self):
         """search creators, for autocomplete"""
-        payload = {
-            'title': 'title test_creator_filtering',
-            'content_url': 'http://example.com/test_creator_filtering',
-            'description': 'description test_creator_filtering',
-            'creators': ['Alice', 'Bob', 'Carol']
-        }
-        json.loads(str(self.client.get('/api/pulse/nonce/').content, 'utf-8'))
-        self.client.post(
-            '/api/pulse/entries/',
-            data=self.generatePostPayload(data=payload)
-        )
         creatorList = json.loads(
             str(self.client.get(
                 '/api/pulse/creators/?search=A'
@@ -30,6 +19,5 @@ class TestCreators(PulseStaffTestCase):
         for creator in self.creators:
             if creator.name.startswith('A'):
                 db_creators.append(creator.name)
-        db_creators.append('Alice')
 
         self.assertEqual(creatorList, db_creators)
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -26,4 +26,10 @@ class TestCreators(PulseStaffTestCase):
                 '/api/pulse/creators/?search=A'
             ).content, 'utf-8')
         )
-        self.assertEqual(creatorList, ['Alice'])
+        db_creators = []
+        for creator in self.creators:
+            if creator.name.startswith('A'):
+                db_creators.append(creator.name)
+        db_creators.append('Alice')
+
+        self.assertEqual(creatorList, db_creators)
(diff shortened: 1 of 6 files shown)


pulseapi/creators/tests.py
```diff
@@ -26,4 +26,10 @@ def test_creator_filtering(self):
                 '/api/pulse/creators/?search=A'
             ).content, 'utf-8')
         )
-        self.assertEqual(creatorList, ['Alice'])
+        db_creators = []
+        for creator in self.creators:
+            if creator.name.startswith('A'):
+                db_creators.append(creator.name)
+        db_creators.append('Alice')
```
Pomax: why the extra Alice? There should be no need to hardcode anything in a test.

pulseapi/entries/test_views.py
```diff
@@ -186,10 +189,12 @@ def test_post_entry_with_mixed_creators(self):
             '/api/pulse/entries/',
             data=self.generatePostPayload(data=payload)
         )
-        creatorList = json.loads(
+        creator_list = json.loads(
             str(self.client.get('/api/pulse/creators/').content, 'utf-8')
         )
-        self.assertEqual(creatorList, ['Pomax', 'Alan'])
+        db_creator_list = [c.name for c in self.creators]
+        db_creator_list.extend(creators)
```
Pomax: same question as above really: why the extend? does the self-extraction not cover it? (if not: should we simply reload `self` from the db?

pulseapi/entries/test_views.py
```diff
@@ -554,6 +559,29 @@ def test_moderation_toggle_by_staff(self):
...
+        Make sure that the entry serializer contains all the custom data needed.
+        Useful test to make sure our custom fields are tested and not
+        removed during API changes
+        """
+
+        entries = self.entries
+        serialized_entries = EntrySerializer(entries, many=True).data
+
+        for i in range(len(serialized_entries)):
+            entry = entries[i]
```
Pomax: Clearly this works, but is there a documented reason why the ordering of `serialized_entries` and `entries` match? (I don't see us forcing an ordering, so a future model change might break this)

pulseapi/profiles/test_models.py
```diff
@@ -0,0 +1,19 @@
...
+from faker import Factory
+
+fake = Factory.create()
+
+
+class UserProfileFactory(factory.Factory):
+    """Generate UserProfiles for tests"""
+    is_active = True
+    user_bio = factory.LazyAttribute(
+        lambda o: 'user_bio {}'.format(''.join(fake.text(max_nb_chars=130)))
```
Pomax: nit, but can we name this parameter? Like `'user_bio {fakeuser}'.format(fakeuser = ....)`?

pulseapi/tests.py
```diff
@@ -44,11 +48,40 @@ def setup_groups():
...
+
+        test.creators.extend(creators)
+        test.entries.append(entry)
+
+
+def setup_users_with_profiles(test):
+    users = [EmailUserFactory() for i in range(3)]
+    for user in users:
+        user.save()
+        profile = UserProfileFactory(
```
Pomax: hm, this is an interesting problem. Right now we automatically build a user profile when a never-before-seen user logs in through gauth, do we want to add to that so that profiles are also build when a new user is built outside of social authentication?

pulseapi/entries/test_views.py
gideonthomas: It matches because we use `entries` to populate `serialized_entries` (passed as the first param to the serializer). So whatever order `entries` is in, `serialized_entries` will follow (regardless of model ordering changes).

pulseapi/tests.py
gideonthomas: probably not needed.

PR #217, opened 2017-09-27
Users can only be associated with one profile

### LINKED ISSUE #214: Figure out how org profiles are created via admin
Branching from https://github.com/mozilla/network-pulse-api/pull/212#issue-259916418, it looks like a user can "own" multiple profiles. I think we did this so that we could allow org profiles to be created without there being a corresponding user.

The problem is that since a user can have multiple profiles, it is hard to precisely pick out which of those profiles represents the user itself (i.e. is not an org profile). There is an `is_group` checkbox on profiles that seems to tell us if it is an org profile or not. Is that what we use to tell if the profile is an org profile or not? 

So, if user1 has profiles A, B, and C, profile B represents the user itself and the `is_group` checkbox is unchecked for that profile, is it safe to assume that profiles A and C are org profiles and their `is_group` checkbox is checked? Since this `is_group` toggle is controlled via admin, it seems risky to rely on profile creators to make sure that for their user, only 1 profile has `is_group` unchecked and all other profiles have it checked.

One thing to note is that no one actually needs to own a profile. The way our model is structured, a profile may or may not be associated with a user. This allows users to create org profiles and the way we can distinguish between org profiles and the user's profile is that the org profiles do not have a user attached to them while the user's profile is attached to the user.

Clarifications that can help fix this issue:
- What is `is_group` used to indicate?
- Do we care about who created an org profile?
- Can we rely on admins to make sure that out of all the profiles they create, only one has `is_group` unchecked viz. their profile?

gideonthomas:
This makes sure that a user can have only one profile (previously, a profile had a user as a foreign key indicating that users can have multiple profiles).

The two biggest changes are:
1. We change the relationship between users and profile to be one-to-one vs. many-to-one (ForeignKey).
2. The relationship field (OneToOne) now exists on the user vs. the profile. This allows us to easily ensure that all users have a profile but not every profile needs a user.

I had to change a bunch of earlier migrations because of this. Some of the migrations did not correctly depend on other migrations that needed to be in place. I also needed to change some of the migrations to correctly import the models they needed.

Another small change that might be noticeable is that while you can find a user's profile using `user.profile`, the reverse relation is `profile.related_user` instead to `profile.user`. I had to name it differently in order to generate the schema migration since we can't have the related_name of the old field in profiles model be the same as the field we are adding to the users model (as they exist in tandem at some point in the migration). I added one workaround for this - I added a property on the profile model called `user`. So, in most places, we can user `profile.user` to access the user object.

Two small changes in terms of the API data served:
- First, I removed the `is_group` boolean from the API data sent back as it is meant to be controlled only via the admin and the front-end does not currently use it.
- Second, the `published_by` attribute previously incorrectly used the user's name vs. the user's profile name (which decides whether to use the custom name or not). It now uses the profile name instead. I made the same change to the public profile serializer so that `name` is always the final name that should be shown (this won't have any implications on the front-end except for making their "is name null" check unnecessary).

Fixes #214


pulseapi/profiles/admin.py
```diff
@@ -42,10 +42,16 @@ class UserProfileAdmin(admin.ModelAdmin):
         'bookmark_count',
     )
 
+    def name(self, instance):
+        return instance.name or '-'
```
Pomax: given that this is the admin view only, should this be 'name or email', rather than "name or -"?

gideonthomas: yep, we can do that

Issue #180, opened 2017-08-15
make Entry auto-split tags with commas, and flag bad tags in Tag .save


Pomax:
We should make sure that when saving an Entry, we check all tags for commas, and clean things up before the Entry is saved, so that Tag instances do the right thing.

The Tag model itself should probably also check for commas when they save just so we can cactch when things go wrong.

xmatthewx:
@Pomax - is this a thing you could poke at in between PTOs

Pomax:
that did not happen, can we put this in the upcoming milestone?

PR #220, opened 2017-09-29
Ensure that tags with commas get separated into real tags

Pomax:
This PR introduces fixes in two places:

1.  the Tag `save` definition checks to see if it contains commas, and if so rewrites its own content such that a tag for name `a,b,c` gets committed as a tag for name `a`, with two additional tags getting `get_or_create`d for 'b' and 'c'.
2. The Entry view that is used by DRF (and acts as our POST route handler) rewrites incoming tags so that a tag array where one or more tags contain commas (like `['test1', 'test2', 'test1,test2,test3']`) gets flattened (in the previous case, to `['test1','test2','test3']`)

This covers the following use cases:

1. creating Tags in the Django/Mezzanine admin view (tag with comma leads to multiple tags)
2. creating Entries in the Django/Mezzanine admin view (tags can only be picked from a multiselect box, not through freeform input)
3. creating Entries through a POST call to api/pulse/entries (tag data, if present, gets rewritten)

This fixes https://github.com/mozilla/network-pulse-api/issues/180

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -428,17 +428,19 @@ class EntriesListView(ListCreateAPIView):
 
             # we also want to make sure that tags are properly split
             # on commas, in case we get e.g. ['a', 'b' 'c,d']
-            print(request_data)
-            tags = request_data['tags'] if 'tags' in request_data else []
-            print(tags)
-            filtered_tags = []
-            for tag in tags:
-                if ',' in tag:
-                    filtered_tags = filtered_tags + tag.split(',')
-                else:
-                    filtered_tags.append(tag)
-            print(filtered_tags)
-            request_data['tags'] = list(set(filtered_tags))
+            if 'tags' in request_data:
+                print(request_data['tags'])
+                tags = request.POST.getlist('tags')
+                print('tags', tags)
+                filtered_tags = []
+                for tag in tags:
+                    print('tag', tag)
+                    if ',' in tag:
+                        filtered_tags = filtered_tags + tag.split(',')
+                    else:
+                        filtered_tags.append(tag)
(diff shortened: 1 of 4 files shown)


pulseapi/utility/userpermissions.py
```diff
@@ -11,17 +11,18 @@ def is_staff_address(email):
     if email is None:
         return False
 
-    parts = email.split('@')
-    domain = parts[1]
+    if '@' in email:
```
Pomax: The old code crashed Django when I tried to test with `admin/admin` rather than `admin@example.com/admin`, so this fix is part of this PR in the sense that without it, you can't log into the admin for testing using a simple local account

pulseapi/entries/views.py
```diff
@@ -426,6 +426,18 @@ def post(self, request, *args, **kwargs):
             # relation with a Through class, so that needs manual labour:
             creator_data = request_data.pop('creators', None)
 
+            # we also want to make sure that tags are properly split
+            # on commas, in case we get e.g. ['a', 'b' 'c,d']
+            if 'tags' in request_data:
```
gideonthomas: Question about this: why do we need it if we're handling this at the model save level? Wouldn't the model save function already handle this when the serializer saves?

Pomax: the model save "makes more tags", but has no way to return them for correct crosslinking to "an entry" (the Tag model strictly speaking has no idea what an Entry is), so we need this in two places - in the Tag.save for instances where someone creates a tag directly, and in the POST acceptor for entries so that we can guarantee we start with a preprocessed tag list, so that every tag ends up crosslinked to the to-be-made-right-now Entry.

PR #222, opened 2017-10-03
Allow posters to indicate they are a content creator

### LINKED ISSUE #160: Setup API for 'Submitter is Creator' feature
When User A submits an item, we want to give them a checkbox to identify themselves as one of the creators. If the form knows who they are and has their profile ID, does it send that along with others? Or, does it tell the api to handle it?

Note: in the future, project creators will be linked to Profiles, not Users. 

Frontend form work here: https://github.com/mozilla/network-pulse/issues/573

Pomax:
This adds an optional `posted_by_creator` bool to the Entry model so that a posting user can be marked as creator for an entry, without having to specify explicit Creator fields. This PR includes the necessary docs update, as well as new test to verify passing in the relevant value as the desired effect during as result of a POST operation.

This closes #160


pulseapi/entries/views.py
```diff
@@ -462,11 +462,11 @@ def post(self, request, *args, **kwargs):
                 )
 
                 # create entry/creator intermediaries
-                if creator_data is not None:
+                if saved_entry.posted_by_creator:
+                    creator_data.append(user.name)
```
gideonthomas: based on the slack convo, my guess is that atm we don't care about the order in which we add the submitter as a creator?

pulseapi/entries/models.py
```diff
@@ -78,6 +78,7 @@ class Entry(models.Model):
     interest = models.CharField(max_length=600, blank=True)
     featured = models.BooleanField(default=False)
     internal_notes = models.TextField(blank=True)
+    posted_by_creator = models.BooleanField(default=False)
```
gideonthomas: nit: maybe use the same terminology we use for the person posting - `published_by`, so it would be `published_by_creator`?

pulseapi/entries/views.py
Pomax: yeah. Although I did just file #223 so we can revisit after MVP

pulseapi/entries/models.py
Pomax: fair - let me update the code

gideonthomas:
r+

xmatthewx:
Can this be merged and closed?

Pomax:
This needs a look at the migrations, because it seems then don't apply cleanly right now

PR #235, opened 2017-10-13
Fix entry-creator migrations to use dynamic models

gideonthomas:
We referred to these models as they are currently in the code vs. using what is in the db at the time of the migration.

I tested this change and it seems to work fine. I also tested it over production data to make sure that it applies cleanly without any data loss.

### DIFF
--- a/pulseapi/entries/migrations/0015_copy_creators_to_ordered_creators.py
+++ b/pulseapi/entries/migrations/0015_copy_creators_to_ordered_creators.py
@@ -1,6 +1,4 @@
 from django.db import migrations
-from pulseapi.entries.models import Entry
-from pulseapi.creators.models import Creator, OrderedCreatorRecord
 
 
 def migrate_creator_data(apps, schema_editor):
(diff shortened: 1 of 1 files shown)


alanmoo:
Talk me through this a bit more. We're changing a migration that already landed on...staging? But has it landed on production? Why are we doing this?

gideonthomas:
Because production breaks when we apply this migration. 
tldr; migrations should always use models as they are in the db at the time of the migration and not as they are in the code.

On staging, since we applied migrations as soon as we landed them, using the models from the code directly wasn't a problem because they would represent what's in the db. On production however, we are applying multiple migrations at a time on the db and the state of the db is not in sync with the models in the code (the migrations are meant to bring them in sync) and in fact, they are very out of sync with the models in the code. So, we can't use the models in the code in a migration and try to apply that migration because the model in the code has properties that are not yet in the db.

alanmoo:
Got it. So to test this I should back up the model to whatever point prod is at and then run all the migrations at once?

PR #237, opened 2017-10-17
Show created entries on profile

### LINKED ISSUE #234: Add created entries to /profiles/<id>
As per https://github.com/mozilla/network-pulse/issues/742#issuecomment-335670891, we need to add created entries to the public profile view.

Changes need to be made to:
https://github.com/mozilla/network-pulse-api/blob/b2f12cfd19e66523dcf3fcb6d79c65032aa8edcc/pulseapi/profiles/serializers.py#L81-L100

alanmoo:
Closes #234

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -100,7 +100,7 @@ class UserProfilePublicSerializer(UserProfileSerializer):
 
     def get_created_entries(self, instance):
         entry_creator_records = OrderedCreatorRecord.objects.filter(creator__profile=instance)
-        return list(map(lambda x: EntrySerializer(x.entry).data, entry_creator_records))
+        return [EntrySerializer(x.entry).data for x in entry_creator_records]
 
     my_profile = serializers.SerializerMethodField()
 
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -84,8 +84,8 @@ class UserProfilePublicSerializer(UserProfileSerializer):
     Serializes a user profile for public view
     """
     name = serializers.CharField(read_only=True)
+
     published_entries = serializers.SerializerMethodField()
-    created_entries = serializers.SerializerMethodField()
 
     def get_published_entries(self, instance):
         user = instance.user
@@ -96,15 +96,13 @@ class UserProfilePublicSerializer(UserProfileSerializer):
             many=True,
         ).data if user else []
 
-    my_profile = serializers.SerializerMethodField()
+    created_entries = serializers.SerializerMethodField()
 
     def get_created_entries(self, instance):
         entry_creator_records = OrderedCreatorRecord.objects.filter(creator__profile=instance)
-        entries = []
-        for entry_creator in entry_creator_records:
-            entry = EntrySerializer(entry_creator.entry).data
-            entries.append(entry)
-        return entries
+        return list(map(lambda x: EntrySerializer(x.entry).data, entry_creator_records))
+
+    my_profile = serializers.SerializerMethodField()
 
     def get_my_profile(self, instance):
         return self.context.get('request').user == instance.user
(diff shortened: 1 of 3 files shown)


pulseapi/profiles/test_views.py
```diff
@@ -0,0 +1,24 @@
...
+        self.assertEqual(response.status_code, 200)
+
+    def test_profile_data_serialization(self):
+        """
+        Make sure profiles have "created_entries" array
+        """
+        id = self.users_with_profiles[0].id
+        response = self.client.get(reverse('profile', kwargs={'pk': id}))
+        entriesjson = json.loads(str(response.content, 'utf-8'))
+        self.assertGreater(len(entriesjson['created_entries']), 0)
```
gideonthomas: this is risky since the len can be an arbitrary value and still pass. We should assert that what is being passed back from the route is what's actually in the db.

Try:
```py
created_entries = []
entry_creators = OrderedCreatorRecord.objects.filter(creator__profile=self.users_with_profiles[0])
for entry_creator in entry_creators:
    created_entries.append(EntrySerializer(entry_creator.entry).data)

self.assertEqual(entriesjson['created_entries'], created_entries)
```

pulseapi/profiles/serializers.py
```diff
@@ -84,6 +85,7 @@ class UserProfilePublicSerializer(UserProfileSerializer):
     """
     name = serializers.CharField(read_only=True)
     published_entries = serializers.SerializerMethodField()
+    created_entries = serializers.SerializerMethodField()
```
Pomax: Can we keep the property and its `def` paired in the code? (maybe move this to above L101 instead, so we have published+def, profile, created+def)

pulseapi/profiles/serializers.py
```diff
@@ -96,5 +98,13 @@ def get_published_entries(self, instance):
 
     my_profile = serializers.SerializerMethodField()
 
+    def get_created_entries(self, instance):
+        entry_creator_records = OrderedCreatorRecord.objects.filter(creator__profile=instance)
+        entries = []
```
Pomax: Python has ways.
```
entries = list(map(lambda x: EntrySerializer(x.entry).data, entry_creator_records))
```

pulseapi/profiles/test_views.py
```diff
@@ -0,0 +1,34 @@
...
+        Make sure profiles have "created_entries" array
+        """
+        id = self.users_with_profiles[0].id
+        response = self.client.get(reverse('profile', kwargs={'pk': id}))
+        entriesjson = json.loads(str(response.content, 'utf-8'))
+
+        created_entries = []
+        entry_creators = OrderedCreatorRecord.objects.filter(
+            creator__profile=self.users_with_profiles[0].id
+            )
```
Pomax: rogue indent

pulseapi/profiles/test_views.py
```diff
@@ -0,0 +1,34 @@
...
+        for entry_creator in entry_creators:
+            created_entries.append(EntrySerializer(entry_creator.entry).data)
```
Pomax: You probably want the same thing here as above:
```
created_entries  = list(map(lambda x: EntrySerializer(x.entry).data, entry_creators))
```
(turning that lambda into its own named var is then probably also worth it)

pulseapi/profiles/serializers.py
alanmoo: This works, but you wanna give me the tl;dr (the documentation) on what it's doing?

pulseapi/profiles/test_views.py
alanmoo: I blame flake.

pulseapi/profiles/test_views.py
alanmoo: Different files though...still worth changing?

Issue #233, opened 2017-10-10
`is_bookmarked` regression - user's bookmarked entry has `is_bookmarked` set to false


mmmavis:
STR

1. make sure you are logged in
2. visit the `/api/pulse/entries/bookmarks/` route to see the list of your bookmarked entries
3. check the `is_bookmarked` property in each Entry object
4. on localhost/staging all of them are set to `false` (should be `true`). On production `is_bookmarked` are properly set to `true`

https://api.mozillapulse.org/api/pulse/entries/bookmarks/
https://network-pulse-api-staging.herokuapp.com/api/pulse/entries/bookmarks/

xmatthewx:
Is this worth discussing on tomorrow's stand? cc @alanmoo to help us remember.

Pomax:
Hm, the fact that it landed on staging suggest the tests for bookmarking still pass, so we may need to do a bit of digging @gideonthomas

gideonthomas:
@Pomax sorry, I fix this in my current patch which isn't on staging yet
Basically it's just how we're accessing `request` from the `context` in the serializer that's wrong.

PR #238, opened 2017-10-17
POST autocompleted creators for entries

gideonthomas:
This PR does two things:
1. Adds functionality to auto-create creator objects when profiles are created and creates creator objects for existing profiles.
2. Changes POST /entries to accept existing creators and create new ones if they don't exist.

I also now force JSON content-type for entries and as a result had to fix some tests for that.

Another small change I made was how we access `request` from the context...we previously accessed it as if it were an object but the context is actually a dict...that's what caused https://github.com/mozilla/network-pulse-api/issues/233.

### To test that creators have profiles:
1. Check the db for a profile you like
2. Run `python manage.py migrate`
3. Check the db again and make sure the `creators_creator` table has an entry with `profile_id` set to your selected profile's id

### To test posting related_creators:
1. Go to `test.example.com:8000/api/pulse/entries/` and go to the bottom of the page
2. Open your console and type in `var x = new XMLHttpRequest(); x.open('GET', 'http://test.example.com:8000/api/pulse/nonce'); x.onreadystatechange = function() { if(x.readyState === 4) { console.log(x.responseText); } }; x.send();` and hit Enter
3. Copy and paste the `nonce` and `csrf_token` into the empty JSON payload auto-generated for you at the bottom of the page. Change `csrf_token` to `csrfmiddlewaretoken`.
4. Fill out at least title. Set `related_creators` property to an array of objects with either an `id` key or a `name` key. The `id` key should be an existing creator `id` (check from the admin). Hit Send.
5. Make sure it succeeds and verify the actual entry.


Pomax:
> I also now force JSON content-type for entries and as a result had to fix some tests for that.
 👍

Pomax:
@gideonthomas the "To test posting related_creators" are literally steps that should be covered by a view test, not requiring any manual work. Did we add that? if not, let's make sure we have that added in this PR

gideonthomas:
the test is here: https://github.com/mozilla/network-pulse-api/pull/238/files#diff-31c66a35c7be6f9f337b1439ab529a9fR641

pulseapi/creators/serializers.py
```diff
@@ -25,24 +28,55 @@ class Meta:
...
+                code='missing data',
+            )
 
-        return profile.id if profile else None
+        if has_id:
+            try:
+                return Creator.objects.get(id=data['id'])
+            except ObjectDoesNotExist:
+                raise ValidationError(
+                    detail=_('No creator exists for the given id.'),
```
Pomax: can we turn this into `'No creator exists for id {id}.'.format(id=data['id'])`? Errors should let us look things up; if it only says "there is an error" that's not a good enough error =)

pulseapi/entries/serializers.py
```diff
@@ -67,6 +69,7 @@ class EntrySerializer(serializers.ModelSerializer):
         required=False
     )
 
+    # QUEUED FOR DEPRECATION: Use the `related_creators` property instead.
```
Pomax: can you add the issue that we filed for this? (if there is none, can you file it? =)

pulseapi/entries/serializers.py
```diff
@@ -79,6 +82,7 @@ def get_creators(self, instance):
         """
         return [ocr.creator.creator_name for ocr in instance.related_creators.all()]
 
+    # QUEUED FOR DEPRECATION: Use the `related_creators` property instead.
```
Pomax: same as above

pulseapi/entries/test_views.py
```diff
@@ -620,12 +630,38 @@ def test_entry_serializer(self):
...
+            db_entry_creator_count = OrderedCreatorRecord.objects.filter(entry=entry).count()
+            self.assertEqual(len(creators_with_profiles), db_entry_creator_count)
+            self.assertEqual(len(related_creators), db_entry_creator_count)
+
+    def test_post_entry_related_creators(self):
+        """
+        Make sure that we can post related creators with an entry
+        """
+        creator1_id = self.creators[0].id
+        creator2_name = 'Pomax'
```
Pomax: I know it's tempting, but please let's not put "currently employee" names in a test suite that can easily outlive someone's employment. It's... really iffy to have to write a patch that changes that person's name to something else later down the line.

pulseapi/entries/views.py
```diff
@@ -208,11 +209,13 @@ class EntryView(RetrieveAPIView):
     queryset = Entry.objects.public()
     serializer_class = EntrySerializer
     pagination_class = None
+    parser_classes = (JSONParser,)
```
Pomax: As a tuple, just to make it explicit this really is one, it might be wise for us to split this over multiple lines.
```
parser_classes = (
  JSONParser,
)
```
(and then the same note for the next handful of same lines of course)

pulseapi/entries/views.py
```diff
@@ -428,16 +433,19 @@ def post(self, request, *args, **kwargs):
             # we also want to make sure that tags are properly split
             # on commas, in case we get e.g. ['a', 'b' 'c,d']
             if 'tags' in request_data:
-                tags = request.POST.getlist('tags')
```
Pomax: I'm glad to see this go. I didn't like it when I wrote it >_>

pulseapi/entries/views.py
```diff
@@ -460,10 +468,7 @@ def post(self, request, *args, **kwargs):
                     moderation_state=moderation_state
                 )
 
-                # create entry/creator intermediaries
-                if saved_entry.published_by_creator:
-                    creator_data.append(user.name)
-
+                # QUEUED FOR DEPRECATION: Use the `related_creators` property instead.
```
Pomax: worth linking to the relevant issue here, too

pulseapi/profiles/apps.py
```diff
@@ -1,5 +1,10 @@
 from django.apps import AppConfig
+from django.utils.translation import ugettext_lazy as _
```
Pomax: oh god no let's never do this O_O

it has a name. Just use that name. Obfuscating literally makes our own dev work harder.

pulseapi/tests.py
```diff
@@ -82,6 +86,22 @@ def setup_users_with_profiles(test):
...
 
 
+class JSONDefaultClient(Client):
+    """
+    Same as a regular test client except the default content type is 'application/json'
+    for the post method instead of 'multipart/form-data'
+    """
+    def post(self, path, data=None, content_type=CONTENT_TYPE_JSON,
+             follow=False, secure=False, **extra):
+        return super(JSONDefaultClient, self).post(
```
Pomax: it looks like `follow` gets lost in this super call.

README.md
```diff
@@ -147,7 +147,7 @@ POSTing of entries requires sending the following payload object:
   tags: optional array of strings
   issue: optional string, must match value from [GET /issues?format=json]
   help_type: optional string, must match value from [GET /helptypes?format=json]
-  creators: optional array of names of creators for the content linked to
+  related_creators: optional array of objects where each object either has an id or a name. The id should be the id of an existing creator.
```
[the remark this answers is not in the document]
gideonthomas: name shouldn't be unique though. That's why we don't key on name. We always create a new creator if a `name` is provided.

pulseapi/tests.py
gideonthomas: good catch!

Issue #229, opened 2017-10-05
report the user's custom name, rather than real name, if the user has set one


mmmavis:
This requires changes in at least two places:

- [ ] `/userstatus` route (new property `name`)
- [x] `creators` meta returned from `/entries/` route

xmatthewx:
Who should take this @alanmoo?

Pomax:
taken and PR filed. this should unblock @mmmavis a bit

PR #240, opened 2017-10-18
add a 'customname' property to userstatus

Pomax:
closes https://github.com/mozilla/network-pulse-api/issues/229 by adding a `customname` property to the `GET api/pulse/userstatus` route. This value is either missing, if the user has no custom name defined in their profile, or is a string that matches what the user set in their profile.

### DIFF
--- a/templates/users/userstatus.json
+++ b/templates/users/userstatus.json
@@ -1,5 +1,6 @@
 {
 {% if loggedin %}	"username": "{{ username }}",
+{% endif %}{% if customname %}	"customname": "{{ customname }}",
 {% endif %}{% if loggedin %}	"email": "{{ email }}",
 {% endif %}	"loggedin": {% if loggedin %}true{% else %}false
 {% endif %}{% if moderator %},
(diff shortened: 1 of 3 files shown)


templates/users/userstatus.json
```diff
@@ -1,5 +1,6 @@
 {
 {% if loggedin %}	"username": "{{ username }}",
+{% endif %}{% if customname %}	"customname": "{{ customname }}",
 {% endif %}{% if loggedin %}	"email": "{{ email }}",
 {% endif %}	"loggedin": {% if loggedin %}true{% else %}false
 {% endif %}{% if moderator %},
```
mmmavis: What's `templates/users/userstatus.json`  for? Is it for determining what gets sent as response?

mmmavis: just a thought - should we always include `customname` as property?

Pomax: wouldn't make a difference on the JS side though:
```
if (result.customname) {
 ...
}
```
this'll do the right thing for `customname` when it's undefined, false, null, or even `""`

mmmavis: oh yea it was just thought as I'm not a API design expert 😉

Issue #242, opened 2017-10-19
Make creators search case-insensitive


alanmoo:
Example: https://network-pulse-api-staging.herokuapp.com/api/pulse/creators/?name=alan doesn't yield results. It probably should so that autocomplete shows those creators. https://network-pulse-api-staging.herokuapp.com/api/pulse/creators/?name=Alan does show results.

cc @xmatthewx

PR #243, opened 2017-10-19
make creator searching case insensitive

Pomax:
fixes https://github.com/mozilla/network-pulse-api/issues/242


pulseapi/creators/views.py
```diff
@@ -26,8 +26,8 @@ def filter_queryset(self, request, queryset, view):
             return queryset
 
         own_name = Q(name__startswith=search_term)
```
gideonthomas: not this?

Pomax: gah

Issue #247, opened 2017-10-23
Make creator search "contains"


alanmoo:
Let's say I want to mark "Pomax" as the creator of a project. But his profile is `Mike 'Pomax' Kamermans`. Typing in a creator name won't currently show his profile as an option. It should.

Pomax:
Based on some discussion in the chat, the right way to do this is probably two-fold:

- on the API side, do a search for `startswith`, and see if that yields fewer results than fit in a single `page_size` sized result set. If so, redo the search using `contains`.
- on the client, make sure to sort the autocomplete suggestions so that any ordering difference between `startswith` and `contains` at the API are invisible to the user.

PR #250, opened 2017-10-25
added icontains fallback when istartswith<page_size

Pomax:
When `startswith_iexact` fails to yield more results than `page_size`, this PR adds a second search for `contains_iexact`, and tacks that to the end of the "startswith" results, to increase the likelihood of someone finding the creator they were looking for.

closes https://github.com/mozilla/network-pulse-api/issues/247


Pomax:
works again

PR #254, opened 2017-10-26
add cacheback and django-rq for request caching on /api/pulse/entries

cadecairos:
This should only apply the cache to `GET /api/pulse/entries` if:

1. `CACHE_ENBLED=true` and if
2. The request is not by a logged in admin (I think)

Lets pick this apart a bit to make sure it's not insane...


cadecairos:
This patch breaks pagination... looking into that

README.md
```diff
@@ -363,3 +363,9 @@ To migrate data, export JSON from the Google Sheets db, and save it in the root
 In `public/migrate.html`, update the endpoint to be the address of the one you're trying to migrate data into. If it's a local db, leave as is.
 Spin up a server from the `public` folder on port 8080. Log in to your API using Oauth (either the hosted site or `test.example.com:8000` if doing this locally)
 Visit `http://test.example.com:8080/migrate.html`, paste the contents of `massagedData.json`, and submit. It will process the entire array of entries one at a time, POSTing them to the server. Check your developer console and network requests if it doesn't complete after a minute or two.
+
+## Caching
+
+The entries list can be cached to improve request performance. Caching requires a Redis server, and a django-rq worker to be running. 
+
+First, define `REDIS_URL` as the connection string to your Redis server. The worker can be started using the following command `python manage.py rqworker default`. This worker will refresh caches asyncronously. Set `CACHES_ENABLED` to `True` in order to enable request caching on the Django app. By default, the entries list is cached for 60 seconds, but it can be overridden with `CACHE_LIFETIME`.
```
Pomax: `CACHE_ENABLED` rather than `CACHES_ENABLED`?

pulseapi/settings.py
```diff
@@ -22,6 +22,9 @@
 
 environ.Env.read_env(os.path.join(BASE_DIR, '.env'))
 env = environ.Env(
+    CACHE_ENABLED=(bool, False),
+    CACHE_LIFETIME=(int, 60),
```
Pomax: which units is this in? (maybe a # comment after to make that clear?)

alanmoo:
Was #255 to move some of the work from the database to the server, and this one caches the server work for certain routes?

alanmoo:
Are we abandoning this approach?

cadecairos:
I think this specific approach can be abandoned, but I think we still should cache the route.

Maybe something simpler, using django's caching api with memcached.

PR #255, opened 2017-10-27
Prefetch related models when listing entries

cadecairos:
This patch uses prefetch_related() on fields with related models in the `EntriesQuerySet` class

The complexity of database queries reduced from O(8n+3) to O(2n+10). There is a minor trade-off: requests returning a single entry will make 12 queries now, where it made 11 queries before.

Admittedly, this isn't ideal still, but it's an improvement. I don't believe an O(1) implementation is possible right now though. The `UserProfile` class uses an accessor method to dynamically query for the username from the related `EmailUser` class, making it impossible to prefetch. Django also seems to generate a count query for each record. I haven't dug into this count query to see if it can be prevented.

### DIFF
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -56,7 +56,21 @@ class EntryQuerySet(models.query.QuerySet):
             # and so if its query set is checked, it'll crash out due
             # to the absence of the associated ModerationState table.
             approved = ModerationState.objects.get(name='Approved')
-            return self.filter(moderation_state=approved)
+            return self.filter(
+                    moderation_state=approved
+                ).prefetch_related(
+                    'tags'
+                ).prefetch_related(
+                    'issues'
+                ).prefetch_related(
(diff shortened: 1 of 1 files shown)


cadecairos:
> Django also seems to generate a count query for each record. I haven't dug into this count query to see if it can be prevented.

Found it. this is the `bookmark_count` field on the serializer, whose getter runs a query to count bookmarks during serialization.

I wonder if we can just fetch all bookmarks for each entry (as an excluded field in the serializer), and have the getter just calculate `len(bookmarks)` or something..

cadecairos:
I got rid of the count queries by prefetching all the bookmark records for an entry. At first I thought it wasn't possible but I just had to position the field name at the right spot in the prefetch_related argument list

alanmoo:
Interesting. I don't think we even use the bookmark count right now, just whatever the "is this bookmarked by the current user" property is.

PR #257, opened 2017-10-30
Update env vars so review apps work

alanmoo:
Closes https://github.com/mozilla/network-pulse/issues/730

### DIFF
--- a/pulseapi/settings.py
+++ b/pulseapi/settings.py
@@ -25,6 +25,8 @@ env = environ.Env(
     DEBUG=(bool, False),
     USE_S3=(bool, False),
     SSL_PROTECTION=(bool, False),
+    CORS_REGEX_WHITELIST=(tuple, ()),
+    CORS_WHITELIST=(tuple, ()),
 )
 SSL_PROTECTION = env('SSL_PROTECTION')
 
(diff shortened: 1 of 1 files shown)


pulseapi/settings.py
```diff
@@ -25,6 +25,8 @@
     DEBUG=(bool, False),
     USE_S3=(bool, False),
     SSL_PROTECTION=(bool, False),
+    CORS_REGEX_WHITELIST=(tuple, ()),
```
cadecairos: I believe that if you specify this as a tuple, comma delimited strings are automatically converted to the right format, so you shouldn't need to call `split(',')` on the variable below (if you use `env()` not `os.getenv()`)

pulseapi/settings.py
```diff
@@ -165,18 +167,19 @@
...
 
 # we want to restrict API calls to domains we know
-CORS_ORIGIN_ALLOW_ALL = False
+if '*' in env('CORS_WHITELIST'):
+    CORS_ORIGIN_ALLOW_ALL = True
+else:
+    # and we want origin whitelisting
+    CORS_ORIGIN_WHITELIST = os.getenv(
+        'CORS_ORIGIN_WHITELIST',
+        'localhost:3000,localhost:8000,localhost:8080,test.example.com:8000,test.example.com:3000'
```
cadecairos: can't this default be specified in an env file? Then we can just use `env()` to grab the setting from the environment.

pulseapi/settings.py
```diff
@@ -165,18 +167,19 @@
...
+    ).split(',')
+    CORS_ORIGIN_REGEX_WHITELIST = env('CORS_REGEX_WHITELIST')
```
cadecairos: we should look into how this works, we could simple specify a regex matching the review app domains that are created.

pulseapi/settings.py
alanmoo: Ok, I was just copying from the network site settings 😁

alanmoo:
I simplified this PR, just the regex should do the trick now

Issue #260, opened 2017-11-21
Expand search to include Creator names


xmatthewx:
@xmatthewx commented on [Tue Oct 24 2017](https://github.com/mozilla/network-pulse/issues/819)

Search is awesome on entries. Let's expand it to enable a search by creator or published_by names.

Pomax:
Analysis reveals that by adding creator and publisher searching to this route, we increase the number of queries by two orders of magnitude, and we sixfold the time required to complete a search. As such, we may need to rethink how we want to search, and what-for.

Some suggestions would be google-style `thing:keyword`, or an "advanced search" where you can explicitly fill in search terms in dedicated input fields, and then if the search takes longer that "makes sense".

@alanmoo @gideonthomas @xmatthewx thoughts?

xmatthewx:
If necessary, I agree it would be better to pull out advanced search rather than impact all search. Or, we could have a simple toggle to search "Projects" or "Creators".

But is it necessary? Let's huddle up, flip/reframe/invert the problem and goal to make sure we aren't missing a quirky and cool solution.

Pomax:
there's always the concept of putting a selector in front of the search, amazon style, so you have a little pulldown that's autoselected to "entries" (implying title, description), but with "tags", "creators", and "publishers" as additional selectors.

There's a fair bit of approach here, might be a good thing to talk over with more people (including designers) as a brainstorm session.

gideonthomas:
After chatting with a django expert in MoTo, he suggests one of two approaches (and agrees that this is a hard problem to solve using just django):
1. **The crude solution** - Create an indexable column in the entries table that encapsulates all of the data from the related tables and the entries table (probably as a string). We can then use `LIKE` sql queries to search that column for whatever search term we want.
2. **The "proper" solution** - Use elastic search + haystack and move all search functionality to this platform. This will be a significant amount of work to add these two things and the django libraries for both of them are not well maintained.

xmatthewx:
Closing. Thanks for the research and ideas @gideonthomas and @Pomax. I've designed an alt approach with tabbed results to separate types of search.

PR #261, opened 2017-12-04
add creator searching back to general entry searching

Pomax:
this addresses 50% of the requirements outlined in https://github.com/mozilla/network-pulse-api/issues/260


Pomax:
closing due to this making search unacceptably slow

PR #262, opened 2017-12-04
Publisher searching

Pomax:
based on top of #261, this addresses the other 50% of the requirements outlined in #260

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -329,7 +329,7 @@ class EntriesListView(ListCreateAPIView):
         'get_involved',
         'interest',
         'related_creators__creator__name',
-        'published_by__nam'e,
+        'published_by__name',
         'tags__name',
     )
     serializer_class = EntrySerializer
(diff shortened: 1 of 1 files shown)


pulseapi/entries/views.py
```diff
@@ -328,6 +328,8 @@ class EntriesListView(ListCreateAPIView):
         'description',
         'get_involved',
         'interest',
+        'related_creators__creator__name',
```
Pomax: this line is introduced by #261

pulseapi/entries/views.py
```diff
@@ -328,6 +328,8 @@ class EntriesListView(ListCreateAPIView):
...
+        'published_by__nam'e,
```
gideonthomas: typo

Pomax:
@gideonthomas fixed that typo

pulseapi/entries/views.py
```diff
@@ -328,6 +328,8 @@ class EntriesListView(ListCreateAPIView):
         'description',
         'get_involved',
         'interest',
+        'related_creators__creator__name',
+        'published_by__name',
```
gideonthomas: hmm, I wonder if it would be better for this to be `published_by__profile__name` (not sure if it would work since `name` isn't an actual model field but is a property on the class)

Pomax:
closing due to this making search unacceptably slow

PR #264, opened 2017-12-07
Making debugging a lot easier

Pomax:
This adds the "Django Debug Toolbar" when running the API server on localhost/127.0.0.1, which makes things _so_ much easier.

image

image


Pomax:
flagging for re-review just to make sure @cadecairos

Issue #241, opened 2017-10-19
Remove deprecated properties from the API


gideonthomas:
We have been adding new properties such as `creator_with_profiles` followed by `related_creator` instead of replacing existing ones in the serializers. Once the front-end and back-end are in sync, we should remove these properties to clean up the code.

gideonthomas:
@alanmoo looks like this still needs to be removed:
https://github.com/mozilla/network-pulse-api/blob/e7127dcbc83896d5852055b03d5ef975b31605c4/pulseapi/entries/views.py#L479-L488

alanmoo:
Good catch, I thought I searched through the repo when cleaning it up pre-Mozfest.

ppatel221:
Hello, I'm a student currently enrolled in an open source class. Is it okay if I work on this?

alanmoo:
Sure thing!

ppatel221:
Ive submitted a PR. I've deleted the lines you specified. But the checks are failing on Travis. https://github.com/mozilla/network-pulse-api/pull/266

alanmoo:
Looks like you’ve got some formatting problems, try running flake8 locally and it should tell you what’s wrong.

-Alan Mooiman,
Mozilla Foundation Engineering Manager
Eastern Standard Time (USA) UTC-5

ppatel221:
I've ran Flake8 and i fixed the formatting errors and deleted some unused variables. It shows me these two errors still. 
**I went from:**
pulseapi/entries/views.py:16:1: F401 'pulseapi.creators.models.OrderedCreatorRecord' imported but unused
pulseapi/entries/views.py:16:1: F401 'pulseapi.creators.models.Creator' imported but unused
pulseapi/entries/views.py:258:17: E722 do not use bare except'
pulseapi/entries/views.py:434:13: E722 do not use bare except'
pulseapi/entries/views.py:439:13: F841 local variable 'creator_data' is assigned to but never used
pulseapi/entries/views.py:480:17: E303 too many blank lines (2)

**To:**
pulseapi/entries/views.py:257:17: E722 do not use bare except'
pulseapi/entries/views.py:433:13: E722 do not use bare except'

What should i do with these excepts

ppatel221:
I just fixed all of the flake8 errors I was getting. After fixing it I submitted a PR and it still won't pass the test. https://github.com/mozilla/network-pulse-api/pull/267

PR #267, opened 2018-01-01
Ran flake8

ppatel221:
Fixes https://github.com/mozilla/network-pulse-api/issues/241

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -13,7 +13,6 @@ from rest_framework.pagination import PageNumberPagination
 from rest_framework.response import Response
 from rest_framework.parsers import JSONParser
 
-from pulseapi.creators.models import Creator, OrderedCreatorRecord
 from pulseapi.entries.models import Entry, ModerationState
 from pulseapi.entries.serializers import EntrySerializer, ModerationStateSerializer
 from pulseapi.profiles.models import UserBookmarks
@@ -434,10 +433,6 @@ class EntriesListView(ListCreateAPIView):
             except:
                 pass
 
-            # we need to split out creators, because it's a many-to-many
-            # relation with a Through class, so that needs manual labour:
-            creator_data = request_data.pop('creators', [])
-
             # we also want to make sure that tags are properly split
             # on commas, in case we get e.g. ['a', 'b' 'c,d']
             if 'tags' in request_data:
@@ -475,8 +470,6 @@ class EntriesListView(ListCreateAPIView):
                     featured=False,
                     moderation_state=moderation_state
                 )
-
-
                 return Response({'status': 'submitted', 'id': saved_entry.id})
             else:
                 return Response(
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -254,7 +254,7 @@ class BookmarkedEntries(ListAPIView):
                 # find the entry for this id
                 try:
                     entry = Entry.objects.get(id=id)
-                except:
+                except entry.DoesNotExist:
                     return
 
                 # find out if there is already a {user,entry,(timestamp)} triple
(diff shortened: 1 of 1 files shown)


alanmoo:
Let's diagnose this in here. Keep in mind that as you update your branch, it will update this PR when you push to that branch, so you don't have to open new PRs every time. Also, please edit your initial comment so it says "Fixes #241" so it's easier to track.

As for the errors themselves, it looks like your tests are failing. Try running `python manage.py test` locally to get more clarity on them. The `except` issue might be from using `except` outside of a `try` block. Learn about those [here](https://www.pythoncentral.io/catching-python-exceptions-the-try-except-else-keywords/)

pulseapi/entries/views.py
```diff
@@ -431,13 +430,9 @@ def post(self, request, *args, **kwargs):
                     encdata = thumbnail['base64']
                     proxy = ContentFile(base64.b64decode(encdata), name=name)
                     request_data['thumbnail'] = proxy
-            except:
+            except thumbnail.DoesNotExist:
```
gideonthomas: why are we catching this exception specifically (which isn't a valid exception thrown here btw)? I don't think we care about what kind of error it is because they are all Bad request errors. I think instead, we should just log the exception and respond with a bad request error instead of `return/pass`

pulseapi/entries/views.py
```diff
@@ -431,13 +430,9 @@ def post(self, request, *args, **kwargs):
...
                 pass
 
-            # we need to split out creators, because it's a many-to-many
-            # relation with a Through class, so that needs manual labour:
-            creator_data = request_data.pop('creators', [])
```
gideonthomas: we definitely have tests that specifically test for this functionality though so you would also need to remove those tests.

ppatel221: The test for this exists in the tests.py file? I'm assuming its this function def setup_entries(test, creator_users).

gideonthomas: these are the tests you'll likely need to modify:
https://github.com/mozilla/network-pulse-api/blob/e83bec051fd17098118e1e901ce60cb4db716d7b/pulseapi/entries/test_views.py#L95-L115

https://github.com/mozilla/network-pulse-api/blob/e83bec051fd17098118e1e901ce60cb4db716d7b/pulseapi/entries/test_views.py#L117-L150

https://github.com/mozilla/network-pulse-api/blob/e83bec051fd17098118e1e901ce60cb4db716d7b/pulseapi/entries/test_views.py#L204-L230

https://github.com/mozilla/network-pulse-api/blob/e83bec051fd17098118e1e901ce60cb4db716d7b/pulseapi/entries/test_views.py#L759-L803

pulseapi/entries/views.py
ppatel221: Ive changed the except's to 
```
except Exception as err:
                logging.error("No such entry",exc_info=err)
```
After making the changes in these two except's. I ran flake out and saw some errors in other files. Should I go ahead and make those changes as well?

gideonthomas: what are the other errors?

ppatel221: Majority of them are:
```c
E501 line too long (80 > 79 characters)
```
Is the way i logged the errors correct?

pulseapi/entries/views.py
ppatel221: I should modify these 4 tests. Not remove them completely right?

gideonthomas: yes, modify them

pulseapi/entries/views.py
gideonthomas: if it is in code that you've changed, yes, modify it so that this error goes away.

Pomax:
It looks like this PR made it into the codebase in some way already, as there are no flake8 errors anywhere in the pulseapi directory anymore.

PR #272, opened 2018-01-18
[WIP] RSS/Atom feeds

### LINKED ISSUE #112: add RSS/Atom feeds
Django has RSS and Atom feed capabilities built in. It would be wise for us to create feeds for the `latest` and `featured` routes by reading through https://docs.djangoproject.com/en/1.11/ref/contrib/syndication/#publishing-atom-and-rss-feeds-in-tandem and simply doing what they tell us to do =)

patjouk:
Fix #112 

This is a WIP PR to add RSS and Atom feeds to Pulse.


Pomax:
Can you explain the architecture here? It feels a little odd to have an API server know what "its client" is, when an API server can technically have a million clients. It might be cleaner, if more work, to handle the RSS as a two-step system:

- any client's own server has an RSS route that asks the API server for tailored base RSS XML that it can rewrite to make work for however it presents entries to the user
- the API server has an RSS XML request route that serves up base XML that can easily be rewritten by consumers to whatever makes sense for them as clients.

That would basically remove the need for the server to know "the" client it has to work with, removing the need for a `PULSE_FRONTEND_HOSTNAME` variable, and then requires a small PR against the client's node server that can fetch the API server's RSS feed and rewrite it with its own paths.

cadecairos:
Do we expect their to be more than one definitive front end client for the pulse entry data? 

IMHO, the frontend and the Django backend should never have been built as two separate projects, and @patjouk and I settled on adding a `PULSE_FRONTEND_HOSTNAME` because of that unfortunate state of the pulse system. 

If it's determined that we indeed expect many clients to have distinct front ends to the data in the Pulse API, I don't think we should even implement an RSS feed on this server at all. The same data is available as JSON, and a client can transform it into RSS on their end.

patjouk:
I will also add that the context of this PR is fixing the pulse bot for slack MozillaFoundation/mofo-devops#555

Pomax:
@cadecairos I still want to write a browser extension for submitting "the page you are on" to Pulse to remove the hassle of having to actually go to a separate website just to submit a pulse entry. That might grow out to an in-browser client (certainly not mvp).

I do agree it would have been nice to have Pulse be a single server setup, but until someone comes up with a way to marry a React application with server-side prerendering with a Python server... we're going to need two different servers =/

patjouk:
@Pomax The bot needs RSS that respect the RSS rfc. 
I will continue to work on this PR for now because we need to fix the bot. If you want to explore doing RSS in the front, can you open another ticket?

patjouk:
So:

**What works:**
- the rss and atom feed for `latest`.
- Same for `featured`, just missing the picture enclosure.
- Tested in a RSS reader + zapier as a bot: works too.

**What could be perfected:**
- Mime type: I used jpeg, which is not always the case >.< I have no idea how to do that dynamically. The only thing that worked was using PIL but it's super heavy.

xmatthewx:
Images are nice, but let's skip or enhance later if mime type gives us too much trouble. We could also add open graph image tag to pulse later and let services pull in thumbs themselves.

README.md
```diff
@@ -343,6 +343,8 @@ The following environment variables are used in this codebase
  - `TOKEN_URI`: optional, defaults to 'https://accounts.google.com/o/oauth2/token' and there is no reason to change it.
  - `SSL_PROTECTION`: Defaults to `False` to make development easier, but if you're deploying you probably want this to be `True`. This sets a slew of security-related variables in `settings.py` that you can override individually if desired.
  Heroku provisions some environmnets on its own, like a `PORT` and `DATABASE_URL` variable, which this codebase will make use of if it sees them, but these values are only really relevant to Heroku deployments and not something you need to mess with for local development purposes.
+ - `PULSE_FRONTEND_HOSTNAME`: Defaults to `localhost:3000`. Used for RSS and Atom feeds: create entries URLs that 
+ links to Pulse website and not the API.
```
cadecairos: Can this read "Used by the RSS and Atom feed views to create entry URLs that link to the network pulse frontend rather than the JSON API"

patjouk: It definitely gives more context. Thanks :)

patjouk:
I applied @cadecairos suggestions and did a bit of cleanup.

Todos after merged:
- [ ] Test in staging.
- [ ] Setup the zapier bot once in prod.

Issue #271, opened 2018-01-17
extend Profile with additional fields


Pomax:
Based on https://github.com/mozilla/network-pulse/issues/657, the Profile model should be extended with several fields that are "standard fields", and several fields that should only be exposed when the "show this profile's extended information" flag is set to true.

General fields for MVP:
- location, as free form text

Extended fields:
- `show_extended` boolean flag to toggle extended field visibility
- `program` (MVP: a single program)
  - `program type` choice selector for known programs
  - `program year` year selector with a reasonably short number of years up to and including whatever year is reasonable but at least the current one
- `affiliation` free form text
- `bio_short` field with 212 character text
- `bio_long` field that takes arbitrary length text content

For the "bio long" we probably want some upper limit to the number of characters, but that's a limit we can iterate on.

**edit** based on https://github.com/mozilla/network-pulse/issues/657 the fields became:

- `profile_type` fixed list of profile types (see 657)
- `show_extended` boolean flag to toggle extended field visibility
- `program` (MVP: a single program)
  - `program type` choice selector for known programs (related-field, admin-curated)
  - `program year` year selector with a reasonably short number of years up to and including whatever year is reasonable but at least the current one (related-field, admin-curated)
- `affiliation` free form text
- `bio_short` field with 212 character text
- `bio_long` field that takes arbitrary length text content

xmatthewx:
Should we call this done?

gideonthomas:
Closing as done. Feel free to re-open if I'm wrong.

PR #274, opened 2018-01-18
Extended profile

Pomax:
This addresses https://github.com/mozilla/network-pulse-api/issues/271 + https://github.com/mozilla/network-pulse/issues/657 by adding the relevant fields, and making them conditionally serialized based on a boolean `enable_extended_information` value in the profile.

When set to False, the profile is serialized as before, when set to True, the additional fields are included in the serialization (in both cases, the `enable_extended_information` flag itself is removed from the serialization).


Pomax:
One thing that's missing: ensuring that PUT operations with extended form data reject that data if the extended profile flag is not set. Also, that the extended profile flag is *never* accepted as form update data. (if it's in the payload, it should get stripped)

Pomax:
added PUT protections

Pomax:
This is missing tests. Tests are needed. Back to work, @pomax

pulseapi/profiles/models.py
```diff
@@ -207,6 +201,89 @@ def thumbnail_image_tag(self):
...
+    # --- extended information ---
+
+    enable_extended_information = models.BooleanField(
+        default=False
+    )
+
+    # currently not captured as Program class,
+    # instead captured as type+year
+
+    PROGRAM_NONE = "None"
```
gideonthomas: any reason we're enforcing choices at the db-level vs. at the serializer/form-level? Doing it at the latter will give us some flexibility.

Pomax: "work", mostly. Doing it at the model level is what I see as the conventional way to do this, but if you have an example of doing this at the Serializer level, I'll happily change it.

gideonthomas: yeah I'm not picky. It is slightly more code since you have to do it to the serializer AND the admin.

Pomax: Can you explain what you mean with that? As far as I can tell, just specifying it in the model means that it's used in building the UI without any need for modifying the Serializer

Pomax:
There are now tests, and making them work revealed that we may have to rewrite all out tests such that we use the `APIRequestFactory`  concept, rather than using test `Client` instances for performing API operations. Doing the latter in fact completely broke testing over unsupported media types and lots of other red herrings.

pulseapi/profiles/models.py
gideonthomas: sorry, my second comment was meant for if you do it at the serializer and admin level instead of at the model level.

pulseapi/profiles/serializers.py
```diff
@@ -24,7 +24,27 @@ class Meta:
...
+    Note that the following fields should only show up when
+    the 'enable_extended_information' flag is set to True:
+
+    - user_bio_long
+    - program_type
+    - program_year
+    - affiliation
     """
+
+    def __init__(self, instance, *args, **kwargs):
```
gideonthomas: `instance` shouldn't be a positional arg since it can be `None` (for e.g. during deserialization).
Here's what the `init` usually looks like for serializer classes: 
https://github.com/encode/django-rest-framework/blob/78367ba1024ec43349aba362758e826fdeafbb03/rest_framework/serializers.py#L111

you can probably just change it to:
```python
if instance is not None:
    if instance.enable_extended_information is False:
        self.fields.pop('user_bio_long')
        self.fields.pop('program_type')
        self.fields.pop('program_year')
        self.fields.pop('affiliation')
    # Whether this flag is set or not, it should not
    # end up in the actual serialized profile data.
    self.fields.pop('enable_extended_information')
```

pulseapi/profiles/test_views.py
```diff
@@ -2,12 +2,20 @@
...
 from django.core.urlresolvers import reverse
 
+from .models import UserProfile
+from .views import UserProfileAPIView
+
 from pulseapi.tests import PulseMemberTestCase
 from pulseapi.entries.serializers import EntrySerializer
 from pulseapi.creators.models import OrderedCreatorRecord
 
+from rest_framework.test import APIRequestFactory, force_authenticate
```
gideonthomas: I don't mind changing to using rest_framework testing tools but it doesn't add too much (only potential request inspection) over the regular django testing tools. For e.g. `force_authenticate` is a small wrapper over `force_login` that django provides and doesn't really change a lot about how it behaves.

pulseapi/profiles/test_views.py
```diff
@@ -30,5 +38,62 @@ def test_profile_data_serialization(self):
...
+        # make sure extended profile data does not show
+        self.assertEqual('program_type' in entriesjson, False)
+
+    def test_extended_profile_data(self):
+        (profile, created) = UserProfile.objects.get_or_create(related_user=self.user)
+        profile.enable_extended_information = True
+        profile.program_type = 'Mozilla Fellow'
+        profile.save()
+
+        profileURL = reverse('profile', kwargs={'pk': profile.id})
```
gideonthomas: let's use python's snakecase vs. camelcase

pulseapi/profiles/urls.py
```diff
@@ -1,11 +1,17 @@
 from django.conf.urls import url
 
 from pulseapi.profiles.views import (
+    UserProfileAPIView,
     UserProfilePublicAPIView,
     UserProfilePublicSelfAPIView,
 )
 
 urlpatterns = [
+    url(
```
gideonthomas: I would suggest that we not create a new route just for update. We have the `/myprofile` route to update the profile so let's stick to using that. Maybe we can add a comment here to explain that create/update/delete are only allowed on your own profile and therefore we have a separate route (without the pk) declared in the main urls.py file.

pulseapi/profiles/views.py
```diff
@@ -13,6 +13,14 @@
 )
 
 
+# utility function for removing keyed values, with key, from a dict
+def removeKey(container, property):
```
gideonthomas: snakecase

pulseapi/profiles/views.py
```diff
@@ -52,15 +60,47 @@ def put(self, request, *args, **kwargs):
...
+        removeKey(payload, 'enable_extended_information')
+
+        '''
+        Do a prefetch on the profile to update. Is the
+        'enable_extended_information' property set to False?
+        In that case, remove any extended properties from the
+        request payload.
+        '''
+
+        try:
```
gideonthomas: this looks like logic that should probably be in the serializer instead. So you can either tack this on to the `init` method (might be complicated) or instead just override the serializer's `to_internal_representation` method.

Pomax: Not sure I agree - this is check is based on a user associated with this specific API call, making the entry point for that call the right place to perform it. The serializer on the other hand should not know about "why" it's being called; it might be due to an HTTP request but it might just as easily be part of a code path that has nothing to do with HTTP interactions, so there is no reason to assume that a serializer has a request object to work with.

gideonthomas: hmm, I don't follow...the serializer would still not know why it is being called because it wouldn't deal with a user at all. All the serializer would care about is its own data. I'm not suggesting we copy/paste this logic as is to the serializer...we would really only need lines 95-99. So the method would look something like this:
```python
def to_internal_value(self, data):
    enable_extended_information = self.fields['enable_extended_information'].get_value(data)
    profile = super(UserProfileSerializer, self).to_internal_value(data)
    if enable_extended_information is not True:
        del profile['user_bio_long']
        del profile['program_type']
        ....
    return profile
```

pulseapi/profiles/models.py
Pomax: In this case I'd say leave it in the model, mostly because the "nothing selected" labels are not strings we're going to be changing a lot. Ideally, we pick ones that work for us, and then never ever touch them again. We will need to update the list of programs and years as years roll over, but that's effectively one migration a year, which feels fine.

pulseapi/profiles/views.py
Pomax: ah, I see.

pulseapi/profiles/views.py
```diff
@@ -13,6 +13,14 @@
 )
 
 
+# utility function for removing keyed values, with key, from a dict
+def remove_key(container, property):
```
gideonthomas: we don't need this here anymore?

Pomax: curious, normally flake8 catches that. removing.

gideonthomas: this won't be something that flake8 would catch because this function is essentially an exported function, so it might be used in some other module and flake8 can't detect that.

Pomax:
Based on https://github.com/mozilla/network-pulse/issues/657 there are some new requirements in terms of this implementation. Invalidating this PR until those are worked in.

Pomax:
Update with relation'd fields:

image

Pomax:
There is something I don't like about the serialization atm, though, maybe @gideonthomas can help: because the profile type, program type, and program year are not foreignkeys, they serialize as the numerical id of the foreign key, rather than the value that key points to.

Is there a clean and simple way to make that be the `record.value` rather than `record.pk`?

Pomax:
In order to bootstrap this to our satisfaction @alanmoo @xmatthewx can we come up with a set of program types that we can automatically insert when we set up the database? Right now we have:

- Profile Types: `plain`, `staff`, `fellow`, `board member`, `grantee`
- Program Type: `tech policy fellowship`, `mozfest speaker`
- Program Year: 2015, 2016, 2017, 2018, 2019

Pomax:
fixed the serialization, it turned out to just need a `StringRelatedKey` type so that was a three line fix (since there were three fields affected)

pulseapi/profiles/test_views.py
```diff
@@ -2,6 +2,8 @@
 
 from django.core.urlresolvers import reverse
 
+from .models import UserProfile, ProgramType
```
gideonthomas: let's add a test for the default saving logic for profiletype

pulseapi/profiles/serializers.py
```diff
@@ -24,12 +33,28 @@ class Meta:
...
     """
     Serializes a user profile.
+
+    Note that the following fields should only show up when
+    the 'enable_extended_information' flag is set to True:
+
+    - user_bio_long
+    - program_type
+    - program_year
+    - affiliation
```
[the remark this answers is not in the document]
Pomax: as I understand it, `profile_type` is a universal field, while the others are not, but maybe @xmatthewx can say whether that's correct or not.

pulseapi/profiles/models.py
```diff
@@ -207,6 +262,54 @@ def thumbnail_image_tag(self):
...
 
+    # --- extended information ---
+
+    enable_extended_information = models.BooleanField(
+        default=False
+    )
+
+    profile_type = models.ForeignKey(
+        'profiles.ProfileType',
+        null=True,
```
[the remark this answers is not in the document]
Pomax: if I remove this, every single test fails O_O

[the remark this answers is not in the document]
alanmoo: o_O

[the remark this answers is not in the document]
gideonthomas: That's interesting...probably some weird bug we're running into, but it's fine, allowing null isn't too bad.

gideonthomas:
Hmm I'm also seeing the search pr logic in this PR...was it merged into this branch?

Pomax:
might have been a result of a rebase

PR #277, opened 2018-01-24
fix review apps

patjouk:
I wanted to use the Review Apps feature from Heroku to test my RSS PR more easily but sadly, it was not working.

The current `app.json` was not properly configured:
screen shot 2018-01-25 at 12 47 28 pm

Since it's a useful feature, here's my attempt to fix this problem :)


patjouk:
As you can see, I struggled a bit to make this PR works 🔥 
My main issue was with the OAuth: since it's a review app, its url will change all the time and it won't be possible to configure OAuth properly.

**What is working right now:**
- the review app: it's starting and running,
- create an admin account using the bash console Heroku or Heroku CLI,
- create entries/stuff in admin works. You can connect the review app API URL to your local dev version of the front and tada, it shows the entries you've just created using the admin.

**What doesn't work right now:**
- I'm having issue understanding how to properly configure the CORS settings: If you use the front to do POST request, you're getting a 403. It's probably fixable once I understand how it works
- Like I already said, it's not possible to log using OAuth.

Pomax:
we *might* be able to make this work with an oauth redirector running on a stable location (say, its own heroku app) but that would be a really brittle solution. For review apps, we should probably add a special env var that says "this is a review instance, automatically grant login if the email domain is in [X,Y,Z]" so we can test with passing and non-passing login addresses

patjouk:
I went for a more lightweight solution: since review apps can do post-deploy script, I wrote a management command that creates a superuser `test@test.org` with a random password. This password is available in the app's deploy logs (`View initial app setup` in the app menu of the pipeline dashboard):

screen shot 2018-02-13 at 12 40 41 pm

If we want to create multiple type of users, we can do the same kind of work we did on the Foundation site with the factories. Right now, I think the review app just needed an easy way to get an admin user.

app.json
```diff
@@ -1,29 +1,46 @@
...
+    "heroku-postgresql:hobby-dev"
   ],
   "buildpacks": [
     {
       "url": "heroku/python"
     }
-  ]
+  ],
+  "scripts": {
+    "postdeploy": "python manage.py review_app_admin"
```
cadecairos: I didn't even know this was a thing we could do! awesome!

pulseapi/users/management/commands/review_app_admin.py
```diff
@@ -0,0 +1,18 @@
...
+from pulseapi.users.models import EmailUser
+
+
+class Command(BaseCommand):
+    help = "Create a super admin user to use on Heroku Review App"
+
+    def handle(self, *args, **options):
+        password = EmailUser.objects.make_random_password()
+        self.stdout.write('Your admin password is: {}'.format(password))
+        user = EmailUser.objects.create_superuser('test', 'test@test.org', password)
```
cadecairos: mind renaming this user to `admin`?

app.json
cadecairos: Could you also add in a section in the README for review apps and how to locate the generated password?

patjouk: Yes, I need to add some docs :)

PR #279, opened 2018-01-30
Profile filtering

Pomax:
builds on top of https://github.com/mozilla/network-pulse-api/pull/274, this PR adds an `api/pulse/profiles/search/` route that can be passed three possible filter arguments:


pulseapi/profiles/migrations/0014_bootstrap_relations.py
```diff
@@ -0,0 +1,36 @@
...
+
+def setup_default_values(apps, schema_editor):
+    ProfileType = apps.get_model('profiles', 'ProfileType')
+    ProfileType.objects.get_or_create(value='plain')
+    ProfileType.objects.get_or_create(value='staff')
+    ProfileType.objects.get_or_create(value='fellow')
+    ProfileType.objects.get_or_create(value='board member')
+    ProfileType.objects.get_or_create(value='grantee')
+
+    ProgramType = apps.get_model('profiles', 'ProgramType')
```
Pomax: note that if we want to bootstrap these properly, we'll need a list from @xmatthewx that can be used as starting point

Pomax:
hm, do we want to also add a migration that upgrades all profiles so that profiles are already set to `plain`, and then extend profiles so that `.create`/`get_or_create` always make sure the default type is `plain`?

Pomax:
superceded by https://github.com/mozilla/network-pulse-api/pull/280

PR #280, opened 2018-02-01
Add a profile searching API route

Pomax:
supercedes #279, which was filed against master

builds on top of https://github.com/mozilla/network-pulse-api/pull/274, this PR adds an `api/pulse/profiles/search/` route that can be passed three possible filter arguments:


pulseapi/profiles/serializers.py
```diff
@@ -45,7 +45,10 @@ class UserProfileSerializer(serializers.ModelSerializer):
 
     def __init__(self, instance=None, *args, **kwargs):
         super().__init__(instance, *args, **kwargs)
-        if instance is not None:
+
+        if instance is not None and type(instance) is UserProfile:
+            # We type-check to prevent this from kicking in for entire
```
gideonthomas: I don't follow this. It would kick in with the result of the queryset object only if it is a UserProfile without this check wouldn't it?

pulseapi/profiles/urls.py
```diff
@@ -20,4 +21,9 @@
     # note that there is also a /myprofile route
     # defined in the root urls.py which connects
     # to the UserProfileAPIView class.
+    url(
```
gideonthomas: nit: can this be a simple `r'^$'` route that uses a list api view defined (mainly don't name it 'search' since it's a view to list profiles?)

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,50 @@ def put(self, request, *args, **kwargs):
...
+        model = UserProfile
+        fields = [
+            'profile_type',
+            'program_type',
+            'program_year',
+        ]
+
+
+class UserProfileAPISearchView(ListCreateAPIView):
+    serializer_class = UserProfileSerializer
```
gideonthomas: I don't know if this is the right serializer to use based on the views above. It looks like this serializer is used on permission protected views and for public views we use `UserProfilePublicSerializer`

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,50 @@ def put(self, request, *args, **kwargs):
...
+
+    filter_backends = (
+        filters.DjangoFilterBackend,
+        filters.SearchFilter,
```
gideonthomas: do we need the `SearchFilter`? I didn't think we were supporting searching on profiles yet...only filtering.

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,50 @@ def put(self, request, *args, **kwargs):
...
+        filters.OrderingFilter,
```
gideonthomas: do we need this? Are we supporting ordering profiles?

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,50 @@ def put(self, request, *args, **kwargs):
...
+        """
+        model = UserProfile
+        fields = [
+            'profile_type',
+            'program_type',
+            'program_year',
+        ]
+
+
+class UserProfileAPISearchView(ListCreateAPIView):
```
gideonthomas: I don't think this should allow `Create` on this view since we don't have any logic to handle that. Can we just have it be a simple list view?
```py
class UserProfilePublicListAPIView(ListAPIView):
    ...
```

pulseapi/profiles/serializers.py
Pomax: tell that to Django: without the typecheck the tests died because `instance` turned out to be an entire QuerySet instead of a single object =/

pulseapi/profiles/urls.py
Pomax: I tried that, but that makes it go into an infinite redirect loop. If you know how to make it not do that, though, an empty route rather than `search` would be much better.

pulseapi/profiles/views.py
Pomax: yarr, good catch

pulseapi/profiles/views.py
Pomax: Hmm, I didn't see any results "not showing up" when calling the API from a non-authenticated session/browser, though.

pulseapi/profiles/views.py
Pomax: ah, copy past error from Entries

pulseapi/profiles/views.py
Pomax: more copy paste error from Entries

pulseapi/profiles/views.py
Pomax: Interestingly, or serializer does not seem to say who the profile belongs to... if I set a custom profile name, that shows, but if I leave it empty, there is literally nothing to tie it to "a human"...

image

probably needs a fix to the Profile serializer in general?

Pomax: Wait a second - why are those fields showing at all, `enable_extended_profile` is false...

Pomax: Okay so this is that weird situation: calling the route through DRF with `/api/pulse/profiles/search?profile_type=plain` for instance, the serializer receives `<QuerySet [<UserProfile: profile for admin@example.com>]>`, rather than that profile.

Pomax:
hm, you'll have to reveal the second "outdated" comment, apparently github won't let me mark it as "Still relevant after PR fixes"

pulseapi/profiles/serializers.py
gideonthomas: oh ok this makes sense since drf dynamically decides whether to implement ListSerializer methods or Serializer methods based on the `many=True` argument passed when creating an instance of the serializer.

pulseapi/profiles/urls.py
gideonthomas: let's land the pr this is dependent on first and then we can try to debug that together.

mmmavis:
Not sure if we wanna fix this here for in a follow-up ticket, but I'm just gonna leave some notes here before I forget 😁 

Sample response 
```json
[
    {
        "custom_name": "",
        "thumbnail": "http://test.example.com:8000/media/images/user-avatars/2018-02-01_232459.4667000000.jpeg",
        "issues": [
            "Online Privacy & Security",
            "Open Innovation"
        ],
        "twitter": "https://example.com/twitter",
        "linkedin": "",
        "github": "",
        "website": "",
        "user_bio": "",
        "profile_type": "fellow",
        "program_type": "senior fellow",
        "program_year": "2016",
        "enable_extended_information": false,
        "affiliation": "",
        "user_bio_long": ""
    }
]
```

- I wonder if we should add `name` to the response. `custom_name` is empty unless the Pulse user has specifically set a `custom_name`. When this happens it makes the `object` in response unidentifiable.
- We'll need to include an `id` or `profile_link` meta in the response so we can use it as the "see work" link. See fellowships site mockup below. "See work" should bring users to the fellow's Pulse profile.

screen shot 2018-02-02 at 2 50 40 pm

gideonthomas:
yeah, I think if we use the `UserProfilePublic` serializer, it should have all of that info

Pomax:
Hm, `UserProfilePublic` still lacks a name. Result with just a superuser:

image

Pomax:
@gideonthomas any idea why the `name` def that we have here, https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/profiles/models.py#L163, does not make it into the serializer here, https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/profiles/serializers.py#L130 ?

gideonthomas:
I do see the name in the screenshot there, it's below `program_year`

pulseapi/profiles/urls.py
Pomax: landed.

Pomax:
I am blind! Already that leaves the question of "how do we use the `/profiles/` route" without getting an infinite redirect

Pomax:
Updated the PR to use `/profiles/` for searching directly, with a special block of code that ensures that if you call that route but without any query arguments that can be filtered on, you get an empty list back, rather than `objects.all()`

Pomax:
closing due to extended-profile landing in master

PR #282, opened 2018-02-02
[Fixes] Expose user ID in `userstatus` endpoint

### LINKED ISSUE #281: Expose user ID in `userstatus` endpoint
To help address https://github.com/mozilla/network-pulse/issues/916, we need an easy way to access the user's id. If a user is logged in, we should add another key to `/userstatus` with their ID.

bhavyaagg:
Signed-off-by: Bhavya Aggarwal bhavyaaggarwal1209@gmail.com

Fixes #281 

Updated the pulseapi/users/views.py and templates/users/userstatus.json files.

### DIFF
--- a/pulseapi/users/test_views.py
+++ b/pulseapi/users/test_views.py
@@ -46,5 +46,6 @@ class TestUserViews(PulseMemberTestCase):
         json_obj = json.loads(string)
         self.assertEqual(json_obj['loggedin'], True)
         self.assertEqual(json_obj['email'], 'test@example.org')
-        self.assertEqual(json_obj['profileid'], '1')
+        self.assertTrue(json_obj['profileid'])
+        self.assertIsInstance(json_obj['profileid'], type("a"))
         self.assertEqual(json_obj['username'], 'plain user')
--- a/pulseapi/users/test_views.py
+++ b/pulseapi/users/test_views.py
@@ -37,7 +37,7 @@ class TestUserViews(PulseMemberTestCase):
     def test_user_status_true(self):
         """
         Assert that an authenticated user calling /userstatus is 200
-        with response { loggedin: true, username: "...", email: "..." }
+        with response { loggedin: true, username: "...", email: "...", profileid: "..." }
         """
         response = self.client.get('/api/pulse/userstatus/')
         self.assertEqual(response.status_code, 200)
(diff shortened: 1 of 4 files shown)


bhavyaagg:
@alanmoo 
Please review whenever possible. 
Also, should i use user.id or user.profile.id. Both are giving the same value when i checked on my local server.
Thank you.

alanmoo:
You should use `profile`. Users were created before profiles, but profiles can exist without users (I believe when you put in a creator that isn't an existing user, a "profile" is created)

bhavyaagg:
@alanmoo I will do that straight away.

bhavyaagg:
@alanmoo I have updated the PR with the required update.

alanmoo:
This seems to work! Can you update `users/test_views.py` to test this?

bhavyaagg:
Sure @alanmoo

bhavyaagg:
@alanmoo I have updated the PR.

bhavyaagg:
@alanmoo 
In local environment, the profileid for the user which is created during tests comes out to be '1' but during the Travis build, it is coming to be '198'.

alanmoo:
That's not surprising, other things might be happening in the database.  It doesn't really matter what the number is though, you probably just want to check that it exists and is the right type.

bhavyaagg:
@alanmoo Thank you for your review.
I have re updated my PR.
Thank you

pulseapi/users/test_views.py
```diff
@@ -46,4 +46,6 @@ def test_user_status_true(self):
         json_obj = json.loads(string)
         self.assertEqual(json_obj['loggedin'], True)
         self.assertEqual(json_obj['email'], 'test@example.org')
+        self.assertTrue(json_obj['profileid'])
+        self.assertIsInstance(json_obj['profileid'], type("a"))
```
alanmoo: I get what this line is doing but it feels odd. Is this some convention I'm not familiar with, or is there a more direct way to check if something is a string in Python? @gideonthomas

bhavyaagg: I think i have found a better way to do this.

bhavyaagg:
@alanmoo I have updated the PR. Thanks a lot for reviewing so fast. :)

alanmoo:
Looks great @bhavyaagg, thanks!

bhavyaagg:
Thank you @alanmoo

PR #284, opened 2018-02-05
Add versioning to the routes

gideonthomas:
I had to override the default url path versioning that drf provides because i think there is a bug in the way it handles optional versions in the url. I filed https://groups.google.com/forum/#!topic/django-rest-framework/r7s9M-VQX_k but haven't seen a response yet. I explained the hack in the actual versioning file.

I also added a `/api/pulse/status/` route that we can use as a healthcheck route (we can file a ticket to make it give more useful info) so that we can have a nice badge in the readme with the current API version listed (I was bored at one point so decided to experiment with it).

~One last thing left to do - add a test to test for versioning.~

## Manual testing steps
Run the server and test these routes in the browser (these have automated tests but manual will also help):
1. `/api/pulse/entries/` - should return entries
2. `/api/pulse/v1/entries/` - should return entries
3. `/api/pulse/v123/entries/` - should return a 404 with a message "Invalid version in url path"
4. `/api/pulse/version1/entries/` - should return a generic 404 not found
5. `/userstatus` - should return user status as json
6. `/v1/userstatus` - should return user status as json
7. `/v123/userstatus` - should return a 404 with a message "Invalid version in url path"
8. `/version1/userstatus` - should return a generic 404 not found
9. `/status/` - should return json with the `latestApiVersion` set to `v1`
10. `/v1/status/` - should return json with the `latestApiVersion` set to `v1`

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,13 +40,13 @@ All API routes are prefixed with `/api/pulse/`. The "pulse" might seem redundant
 
 ## How is the API versioned?
 
-All Pulse API routes are versioned via their url path so that any future changes made to data responses will not break API clients that rely on specific versions of the data. To get a specific version of the API, add the version after the API prefix (`/api/pulse/`) in the url. For e.g. if you want version 2 (v2) of the API for the `entries` route, you would query the `/api/pulse/v2/entries/` url.
+All Pulse API routes are versioned via their url path so that any future changes made to data responses will not break API clients that rely on specific versions of the data. To get a specific version of the API, add the version after the API prefix (`/api/pulse/`) in the url. For example, if you want version 2 (v2) of the API for the `entries` route, you would query the `/api/pulse/v2/entries/` url.
 
-API routes that do not include the `/api/pulse/` prefix are also versioned in the same way. For e.g. `/v1/login/` is a valid versioned route without the prefix. There are two exceptions to this - the `/rss/` prefixed routes and the `/atom/` prefied routes are not versioned.
+API routes that do not include the `/api/pulse/` prefix are also versioned in the same way. For example, `/v1/login/` is a valid versioned route without the prefix. Note however that the `/rss` and `/atom` routes are unversioned, as these are syndication endpoints.
 
 ## What happens if you don't specify an API version in the URL?
 
-To maintain legacy support, if the version is not specified in the url, we default to version 1 (v1) of the API. For e.g. querying `/api/pulse/entries/` will have the same result as querying `/api/pulse/v1/entries/`. However, we strongly recommend specifying a version in the URL as this feature may be removed in the future.
+To maintain legacy support, if the version is not specified in the url, we default to version 1 (v1) of the API. For example, querying `/api/pulse/entries/` will have the same result as querying `/api/pulse/v1/entries/`. However, we strongly recommend specifying a version in the URL as this feature may be removed in the future.
 
 ## Supported API Versions
 
@@ -55,6 +55,8 @@ To maintain legacy support, if the version is not specified in the url, we defau
 ---
 
 # General Routes
+
+All routes, including those with the `/api/pulse/` prefix should include a version in the URL (see the section on [Versioning](#versioning)) and although a URLs without the version are currently supported, versions will be made mandatory in the future.
 
 ## Login routes
 
--- a/README.md
+++ b/README.md
@@ -53,8 +53,6 @@ This is the route that oauth2 login systems must point to in order to complete a
 
 ### `GET /api/pulse/v1/userstatus/`
 
-**Note:** This API route does not have a version in it since it does not rely on data in the database.
-
 This gets the current user's session information in the form of their full name and email address.
 
 The call response is a JSON object of the following form:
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -30,6 +44,8 @@ Please run the server and see [http://localhost:8000/entries](http://localhost:8
 
 ### `GET /api/pulse/nonce/`
 
+**Note:** This API route does not have a version in it since it does not rely on data in the database.
```
Pomax: For ease of development, would it be possible to have all routes accepted the `vX` infix, so that on the client side the API prefix can be kept a single string? (e.g. Set `/api/pulse/v1/` as API root, without having to main two different roots for versioned vs. unversioned?)

README.md
```diff
@@ -14,11 +15,24 @@ This is the REST API server for the Mozilla Network Pulse project.
 - [Migrating data](#migrating-data)
 - [Resetting your database because of incompatible model changes](#resetting-your-database-because-of-incompatible-model-changes)
 
+## API Versioning
+
+Any API route that relies on data from the database is versioned (due to its mutability) via its url path so that any changes made to the data responses will not break API clients that rely on specific versions of the data. To get a specific version of the API, add the version after the API prefix (`/api/pulse/`) in the url. For e.g. if you want version 2 (v2) of the API for the `entries` route, you would query the `/api/pulse/v2/entries/` url.
+
+When a new route is added that relies on data in the database, make sure the url pattern setup for that route calls the appropriate versioning function to generate a versioned route. The verioning functions can be found in [utility/urlutils.py](https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/utility/urlutils.py)
+
+#### What happens if you don't specify an API version in the URL?
```
Pomax: With this explanation provided, it seems better to omit the `/api/pulse/v1` prefixes in the rest of the documentation, and instead point to a pre-v2-update tagged version of this README.md once we do release an API v2

README.md
```diff
@@ -271,12 +289,28 @@ X-CSRFToken: required csrf token string obtained from [GET /nonce]
...
 
+### `GET /api/pulse/status/`
+
+**Note:** This API route does not have a version in it since it does not rely on data in the database.
+
+This is a healthcheck route that can be used to check the status of the pulse API server. The response contains the following information:
+
+```
+{
+  latestApiVersion: <version string>
```
Pomax: if we're adding this, does it make sense to also add a `releaseCommit` that tells us which version of the codebase is running?

pulseapi/entries/views.py
```diff
@@ -22,7 +22,7 @@
 
 
 @api_view(['PUT'])
-def toggle_bookmark(request, entryid):
+def toggle_bookmark(request, entryid, **kwargs):
```
Pomax: can we make this a named argument instead, or does this *have* to be kwargs?

Pomax:
ping @alanmoo on this, too, to figure out what our policy should be here (as our first project with API versioning, this'll set the precedent for other projects, worth making sure what we land is what we want to act as example for others)

alanmoo:
I agree that all the routes should be tagged, regardless of if they're database related or not. Without that it seems like a near-definite "gotcha" for someone in the future.

pulseapi/entries/views.py
gideonthomas: `kwargs` is cleaner otherwise I'll have to add flake8 exceptions for each line since we aren't using the argument anywhere in the function

README.md
gideonthomas: Filed https://github.com/mozilla/network-pulse-api/issues/290

pulseapi/urls.py
```diff
@@ -25,27 +25,31 @@
...
     # API routes
-    url(r'^api/pulse/', include('pulseapi.users.urls')),
-    url(r'^api/pulse/entries/', include('pulseapi.entries.urls')),
-    url(r'^api/pulse/profiles/', include('pulseapi.profiles.urls')),
-    url(r'^api/pulse/tags/', include('pulseapi.tags.urls')),
-    url(r'^api/pulse/issues/', include('pulseapi.issues.urls')),
-    url(r'^api/pulse/helptypes/', include('pulseapi.helptypes.urls')),
-    url(r'^api/pulse/creators/', include('pulseapi.creators.urls')),
+    url(versioned_api_url(), include('pulseapi.users.urls')),
+    url(versioned_api_url(r'entries/'), include('pulseapi.entries.urls')),
```
Pomax: we're using forward slashes, not backslashes, so do we need the raw string prefix here?

pulseapi/users/views.py
```diff
@@ -63,7 +69,7 @@ def new_nonce_value(request):
 
 
 # API ROUTE: /nonce
-def nonce(request):
+def nonce(request, **kwargs):
```
Pomax: what are all these kwargs used for? Future devs might see this, see the function not actually accesss kwargs anywhere, and go "we can remove this"; do things not work without the kwargs argument?

pulseapi/users/views.py
```diff
@@ -265,3 +271,15 @@ def callback(request):
     return HttpResponseNotFound(
         "callback happened without an error or code query argument: this should not be possible."
     )
+
+
+@api_view()
+@renderer_classes((JSONRenderer,))
```
Pomax: @(#%&()@%*@#PYTHON!!  (omg this is the worst syntax)

Pomax:
*mostly

pulseapi/users/views.py
gideonthomas: I'll add a note. We need this function to accept the version (whatever the parameter is called) even though we don't use it in the function, so I don't want to use a named param coz flake8 would complain about unused vars.

pulseapi/urls.py
gideonthomas: nah it's not needed I can remove

PR #285, opened 2018-02-07
Add profile filtering as `/api/pulse/profiles/?...`

Pomax:
supercedes #280, which was filed against a now-merged branch

PR adds an `api/pulse/profiles/` route that can be passed three possible filter arguments:

- searches without any of those types, even if they include other query arguments, will yield an empty set.


pulseapi/profiles/views.py
```diff
@@ -66,3 +67,62 @@ def put(self, request, *args, **kwargs):
...
+    )
+
+    @property
+    def qs(self):
+        """
+        Ensure that if the filter route is called without
+        a legal filtering argument, we return an empty
+        queryset, rather than every profile in existence.
+        """
+        queries = self.request.GET
```
gideonthomas: let's handle the `None` case for `self.request` as well since filters are not required to be tied to requests. we can probably just return `super().qs` if `request` doesn't exist.

pulseapi/profiles/test_views.py
```diff
@@ -105,3 +105,58 @@ def test_program_year_uniqueness(self):
         # as found in the bootstrap migration:
         (profile, created) = ProgramYear.objects.get_or_create(value='2018')
         self.assertEqual(created, False)
+
+    def test_profile_searching(self):
+        profile_types = ['a', 'b', 'c']
+        program_types = ['a', 'b']
+        program_years = ['a', 'b']
+
+        for v1 in profile_types:
```
gideonthomas: any reason we're not doing this in the bootstrap logic?

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,62 @@ def put(self, request, *args, **kwargs):
...
+        """
+        model = UserProfile
+        fields = [
+            'profile_type',
+            'program_type',
+            'program_year',
+        ]
+
+
+class UserProfileAPISearchView(ListAPIView):
```
gideonthomas: let's call this `UserProfileListAPIView`

pulseapi/profiles/urls.py
```diff
@@ -20,4 +21,9 @@
     # note that there is also a /myprofile route
     # defined in the root urls.py which connects
     # to the UserProfileAPIView class.
+    url(
+        r'^$',
+        UserProfileAPISearchView.as_view(),
+        name='profile_search',
```
gideonthomas: `profile_list`

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,62 @@ def put(self, request, *args, **kwargs):
...
+        name='program_type__value',
+        lookup_expr='iexact',
+    )
+    program_year = django_filters.CharFilter(
+        name='program_year__value',
+        lookup_expr='iexact',
+    )
+
+    @property
+    def qs(self):
```
gideonthomas: let's add a test for an invalid query param

pulseapi/profiles/migrations/0015_auto_20180205_1259.py
```diff
@@ -31,16 +31,16 @@ class Migration(migrations.Migration):
         migrations.AlterField(
             model_name='userprofile',
             name='profile_type',
-            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.CASCADE, to='profiles.ProfileType'),
+            field=models.ForeignKey(blank=True, null=True, on_delete=django.db.models.deletion.SET_NULL, to='profiles.ProfileType'),
```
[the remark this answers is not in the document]
Pomax: deleting a profile_type from the list of profile_types should **not** delete all profiles that refer to that now non-existent profile_type. `CASCADE` is a dangerous option. `SET_NULL` just turns the foreignkey reference into a null, without doing anything else.

pulseapi/profiles/views.py
```diff
@@ -66,3 +67,62 @@ def put(self, request, *args, **kwargs):
             pass
 
         return super(UserProfileAPIView, self).put(request, *args, **kwargs)
+
+
+class ProfileCustomFilter(filters.FilterSet):
```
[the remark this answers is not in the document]
Pomax: I'll add this comment as documentation

pulseapi/profiles/test_views.py
Pomax: I wanted this test to verify the functionality irrespective of prepopulated data.

pulseapi/profiles/views.py
Pomax: added.

pulseapi/profiles/migrations/0015_auto_20180205_1259.py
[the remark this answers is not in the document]
Pomax: https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/profiles/migrations/0016_auto_20180201_1257.py already has this migration... I will remove this file.

PR #293, opened 2018-02-13
Add profile_id to the profile JSON response

Pomax:
this adds `profile_id` as JSON field for profiles, with a test to make sure it's there, and the right value.

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -57,10 +57,7 @@ class UserProfileSerializer(serializers.ModelSerializer):
             # end up in the actual serialized profile data.
             self.fields.pop('enable_extended_information')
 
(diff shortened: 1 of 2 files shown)


pulseapi/profiles/serializers.py
```diff
@@ -57,6 +57,11 @@ def __init__(self, instance=None, *args, **kwargs):
             # end up in the actual serialized profile data.
             self.fields.pop('enable_extended_information')
 
+    profile_id = serializers.SerializerMethodField()
```
gideonthomas: probably don't need a `methodfield` for this, you can just do:
```
profile_id = serializers.ReadOnlyField(source='id')
```

Pomax: oh nice, that is even smaller.

PR #294, opened 2018-02-13
switch unused Location class to a location string

Pomax:
this removes the unused Location class, and gives profiles a location string instead, with corresponding update to `test_views`.


Pomax:
I have no idea what's going on here.

Locally:
```
(venv) c:\Users\Mike\Documents\Git\collaborations\mozilla\network-pulse-api>manage test
Creating test database for alias 'default'...
............................................................
----------------------------------------------------------------------
Ran 60 tests in 4.563s

OK
Destroying test database for alias 'default'...
```

but CI keeps failing because of... preexisting data?

@cadecairos @gideonthomas did we set up something weird db-wise for the CI tests to fail on preexisting data, even when there shouldn't be any?

PR #295, opened 2018-02-13
Only allow profile_type to be edited via admin interface

### LINKED ISSUE #286: add admin-restriction for changing a profile's profile type
https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/profiles/serializers.py#L46 currently allows users to change their own profile type and that's not what we want.

This should check if the requesting user has admin privileges, and if not, remove `profile_type` from the list of passed properties.

gideonthomas:
Fix #286 

You can now only edit the profile type via the admin interface.

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -117,6 +117,7 @@ class UserProfileSerializer(serializers.ModelSerializer):
         Meta class. Because
         """
         model = UserProfile
+        read_only_fields = ('profile_type',)
         exclude = [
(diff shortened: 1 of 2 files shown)


pulseapi/profiles/serializers.py
```diff
@@ -117,6 +117,7 @@ class Meta:
         Meta class. Because
         """
         model = UserProfile
+        read_only_fields = ('profile_type',)
```
Pomax: GAH! of course

PR #298, opened 2018-02-15
Procfile update

patjouk:
I encountered something strange will working on failed review apps: sometimes, the `review_app_admin` management command would crash randomly with a psycopg2 error:

```
Traceback (most recent call last):

  File "/app/.heroku/python/lib/python3.6/site-packages/django/db/backends/utils.py", line 64, in execute

    return self.cursor.execute(sql, params)

psycopg2.ProgrammingError: relation "profiles_profiletype" does not exist

LINE 1: ...letype"."id", "profiles_profiletype"."value" FROM "profiles_...
```

After checking Heroku deployment logs, I realized that sometimes, the management command [would be called in the middle of the migrate step](https://gist.github.com/patjouk/9b67f06f9909e2016d976b3da7ec99ab). It was weird because the command is called as a post-deploy step, while the migrate [is normally happening in the release step](https://devcenter.heroku.com/articles/release-phase#review-apps-and-postdeploy-script).
I checked our `Procfile` and realized that release steps and web ones were mixed up in the same script.
I added [gunicorn](https://devcenter.heroku.com/articles/python-gunicorn) so we can experiment and maybe gain some performance. But more importantly, I created a release step that contains the migrate. 

About the `generate_client_secret.py` included in the web step: I need to put it there but I'm sure we can find a better solution. If I put it as in release step, the file get deleted when the dyno is restarted, which crashes the app because it can find that file anymore >.<


Procfile
```diff
@@ -1 +1,2 @@
-web: sh startup.sh
+release: sh release.sh
```
cadecairos: lets just run migrate as the release command

PR #302, opened 2018-02-21
Fix up CORS_ORIGIN_REGEX_WHITELIST parsing so that it's properly set up

cadecairos:
environ doesn't have the ability to load strings as tuples if they contain commas, which it considers a separator.

In order to get CORS regex whitelisting working, I have it loading in a single string from an environment file, which is compiles into a regular expression, then passes to `CORS_ORIGIN_REGEX_WHITELIST` as a tuple.

This is a first-come-first served kind of review. have at it!

### DIFF
--- a/pulseapi/settings.py
+++ b/pulseapi/settings.py
@@ -25,9 +25,10 @@ env = environ.Env(
     DEBUG=(bool, False),
     USE_S3=(bool, False),
     SSL_PROTECTION=(bool, False),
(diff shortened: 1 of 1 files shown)


gideonthomas:
suggested fix: https://gist.github.com/gideonthomas/f275a727d77863f7fce643a29e7f0af8

patjouk:
Closing in favor of #334

PR #303, opened 2018-02-21
Remove hard coded SECRET_KEY

cadecairos:


pulseapi/settings.py
```diff
@@ -28,7 +28,9 @@
     CORS_REGEX_WHITELIST=(tuple, ()),
     HEROKU_APP_NAME=(str, ''),
     PULSE_FRONTEND_HOSTNAME=(str, ''),
+    SECRET_KEY=(str, '')
```
gideonthomas: don't add a default, make it crash without a secret key

cadecairos: this will crash it. the key needs to be at least 32 characters, and '' counts as unset to Django.

gideonthomas:
r+

PR #308, opened 2018-02-26
Significantly improve the admin experience

alanmoo:


pulseapi/profiles/admin.py
```diff
@@ -46,6 +46,24 @@ class UserProfileAdmin(admin.ModelAdmin):
         'bookmark_count',
     )
 
+    list_display = (
+        'name',
+        'user_account',
+        'program_type',
+        '__str__',
```
Pomax: This just adds another column that is a copy of the string that's in the `user_account` column already, so I don't think there's any real value in including this?

PR #314, opened 2018-02-27
Optimize BD queries

patjouk:
I took the opportunity that we now have a fake data generator to revisit our slow queries.

I went from:
screen shot 2018-02-27 at 6 30 42 pm
to:
screen shot 2018-02-27 at 6 31 03 pm

:tada:

### DIFF
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -67,7 +67,7 @@ class EntryQuerySet(models.query.QuerySet):
                     'bookmarked_by__profile__related_user',
                     'published_by__profile',
                     'moderation_state',
-                    'related_creators',
+                    'related_creators__creator',
                 )
 
         except:
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -64,6 +64,7 @@ class EntryQuerySet(models.query.QuerySet):
                     'help_types',
                     'published_by',
                     'bookmarked_by',
+                    'bookmarked_by__profile__related_user',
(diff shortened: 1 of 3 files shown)


pulseapi/entries/models.py
```diff
@@ -64,9 +64,10 @@ def public(self):
                     'help_types',
                     'published_by',
                     'bookmarked_by',
+                    'bookmarked_by__profile__related_user',
                     'published_by__profile',
                     'moderation_state',
-                    'related_creators',
+                    'related_creators__creator__profile__related_user',
```
gideonthomas: this seems like a pretty big join query. Is this faster than getting the fields on-demand in separate queries?

pulseapi/entries/serializers.py
```diff
@@ -109,8 +109,8 @@ def get_is_bookmarked(self, instance):
             user = self.context['request'].user
 
         if user and user.is_authenticated():
-            res = instance.bookmarked_by.filter(profile=user.profile)
-            return res.count() > 0
+            # instance.bookmarked_by.all() is already prefetched and cached in the QuerySet
+            return any(bookmark_by.profile.user == user for bookmark_by in instance.bookmarked_by.all())
```
gideonthomas: nice!

pulseapi/entries/models.py
gideonthomas: to be profiled/verified on prod

gideonthomas:
This is awesome! Thanks @patjouk

PR #318, opened 2018-03-01
Allow choosing creator for a profile in admin

gideonthomas:
This allows you to choose a creator for a profile. This is by no means a final solution (it also has some bad performance issues) but is meant to unblock the fellowships work.

Things to test in the admin:
1. Create a new profile and make sure you can let it automatically create a Creator for you.
2. Create a new profile and make sure you can select an existing Creator to associate with this profile. Check that the previous profile that the existing Creator was associated with is no longer associated with that profile.
3. Update an existing profile and make sure that you can change its creator to another existing Creator. Make sure that the other existing Creator is now no longer associated to the profile it was previously associated with (if it wasn't an orphan profile) and that the previous Creator that was associated with the profile you are updating has been orphaned (it'll say 'no profile' next to the creator name but will have the same name as your profile, i.e. you should see two creators with the same name as your profile, but one will have 'no profile' next to its name).
4. Update an existing profile and make sure that you can change its creator to the option that automatically creates a new creator.

cc @xmatthewx (do the above test cases cover your use case?).

### DIFF
--- a/pulseapi/profiles/static/admin/js/creator-profile-rebind-protection.js
+++ b/pulseapi/profiles/static/admin/js/creator-profile-rebind-protection.js
@@ -1,0 +1,13 @@
+document.addEventListener("DOMContentLoaded", function(e) {
+  var selector = document.getElementById("id_creator");
+  selector.addEventListener("change", function(e) {
+    result = window.confirm(
+      "Are you sure you want to change this binding?\n\n" +
+      "WARNING: If you set this to a creator that is already associated with another profile, that profile will lose its association to the selected creator and all entries that referenced that creator will now link to this profile."
+    );
+
+    if (!result) {
+      e.preventDefault();
+    }
+  });
+});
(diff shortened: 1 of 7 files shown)


Pomax:
This crashes quite hard for me once I assign an existing creator to a profile, then try to load the original profile. STR:

- create a new profile, with automatic creator. Save that profile
- edit that profile and change the creator to an existing user's creator record.
- save and go to the profile list.
- try to open that existing user's profile: django crashes

pulseapi/profiles/static/admin/js/creator-profile-rebind-protection.js
```diff
@@ -0,0 +1,13 @@
+document.addEventListener("DOMContentLoaded", function(e) {
+  var selector = document.getElementById("id_creator");
+  selector.addEventListener("change", function(e) {
+    result = window.confirm(
+      "Are you sure you want to change this binding?\n\n" +
+      "WARNING: If you set this to a creator that is already associated with another profile, that profile will lose its association to the selected creator and all entries that referenced that creator will now link to this profile."
```
Pomax: If we also add "This change CANNOT be undone" then this is perfect.

xmatthewx:
@gideonthomas thanks for advancing this. I'm going to test this now and open a new ticket to discuss follow up. If this works, I think we should push it out to unblock fellow stuff, but we'll have to adjust some things soon.

PR #320, opened 2018-03-05
Add route for profile favorites

### LINKED ISSUE #304: Update profile endpoints for performance
Based on the conversation in #292, let's update the `profiles` endpoints.

- [x] Profile filtering remains in place. Could use some documentation, but that's covered in #185 
- [x] `/profile/{id}` - a specific profile without published/created/fav'd entries
- [x] `/profile/{id}/entries` - a count of all entries related to a user grouped by created/published/favorited
- [x] A specific subset of entries* related to a profile
  - `/profiles/:id/entries?created`
   - `/profiles/:id/entries?published`
   - `/profiles/:id/entries?favorites`
   - `/profiles/:id/entries?published&created` and other combinations

\* Entries returned this way should expose:
  * ID
  * Title
  * Image
  * IsBookmarked
  * related_creators
  * moderation_state (?)
  * created

Anything else should only be exposed if it comes by default without impacting performance
(https://api.mozillapulse.org/api/pulse/profiles/59/ and https://mozillapulse.org/profile/59 is a good reference profile)

gideonthomas:
Fix #304 

This adds a route to get entries associated for a profile based on the querystring - `?created`, `?published`, or `?favorited`. Refer to the linked ticket to see the different routes (getting a profile without these entries is now versioned under `v2/`). This PR also includes optimizations for the queries needed to fetch these entries - the previous queries ran under O(n) + C time where C is a pretty large constant, the new queries run under O(1) time - 8 queries for created and favorited entries and 7 queries for published entries.

If none of the params are passed via querystring, it just returns the number of entries associated with that profile (all created, published, favorited, de-duped).

cc @patjouk this is the work I have so far

### DIFF
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -73,9 +73,10 @@ class EntryQuerySet(models.query.QuerySet):
             'help_types',
             'published_by',
             'bookmarked_by',
+            'bookmarked_by__profile__related_user',
             'published_by__profile',
             'moderation_state',
-            'related_creators',
+            'related_creators__creator__profile__related_user',
         )
 
 
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -47,7 +47,7 @@ class EntryQuerySet(models.query.QuerySet):
 
     def public(self):
         """
-        Return all entries to start
+        Return all entries that have been approved
         """
         try:
             # This has to happen in a try/catch, so that migrations
@@ -56,23 +56,27 @@ class EntryQuerySet(models.query.QuerySet):
             # and so if its query set is checked, it'll crash out due
             # to the absence of the associated ModerationState table.
             approved = ModerationState.objects.get(name='Approved')
-            return self.filter(
-                    moderation_state=approved
-                ).prefetch_related(
-                    'tags',
-                    'issues',
-                    'help_types',
-                    'published_by',
-                    'bookmarked_by',
-                    'bookmarked_by__profile__related_user',
-                    'published_by__profile',
-                    'moderation_state',
-                    'related_creators__creator__profile__related_user',
-                )
+            return self.filter(moderation_state=approved)
 
         except:
             print("could not make use of ModerationState!")
             return self.all()
+
+    def with_related(self):
+        """
+        Return all entries with their related data as separate queries
+        """
+
+        return self.prefetch_related(
+            'tags',
+            'issues',
+            'help_types',
+            'published_by',
+            'bookmarked_by',
+            'published_by__profile',
+            'moderation_state',
+            'related_creators',
+        )
 
 
 class Entry(models.Model):
--- a/pulseapi/settings.py
+++ b/pulseapi/settings.py
@@ -201,6 +201,7 @@ STATIC_ROOT = os.path.join(BASE_DIR, 'static')
 # and <version value> is the value of the version that will be used in URLs
 API_VERSION_LIST = [
     ('version_1', 'v1',),
+    ('version_2', 'v2',),
 ]
 DEFAULT_VERSION = 'version_1'
 # A dictonary of api versions with the value of each version key being
--- a/pulseapi/tests.py
+++ b/pulseapi/tests.py
@@ -164,6 +164,8 @@ class PulseMemberTestCase(TestCase):
     """
     A test case wrapper for "plain users" without any staff or admin rights
     """
+    maxDiff = None
+
     def setUp(self):
         boostrap(
             self,
@@ -179,6 +181,8 @@ class PulseStaffTestCase(TestCase):
     """
     A test case wrapper for "staff" users, due to having a mozilla login
     """
+    maxDiff = None
+
     def setUp(self):
         boostrap(
             self,
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ To maintain legacy support, if the version is not specified in the url, we defau
 
 ## Supported API Versions
 
+- Version 2 - `/api/pulse/v2/` and `/v2/`
 - Version 1 - `/api/pulse/v1/` and `/v1/`
 
 ---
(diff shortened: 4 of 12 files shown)


pulseapi/entries/models.py
```diff
@@ -56,24 +56,28 @@ def public(self):
...
         except:
             print("could not make use of ModerationState!")
             return self.all()
 
+    def with_related(self):
+        """
+        Return all entries with their related data as separate queries
+        """
+
+        return self.prefetch_related(
```
patjouk: Why did you revert the change made in the prefetch?

gideonthomas: Oh I just split the queries into two chainable functions. The prefetch query is the same though. I needed one without the prefetch because it ran too many queries for a query in which I just needed moderated entries that were filtered.

gideonthomas: Oh nevermind, I see what you meant. Must be a rebase error, sorry, fixing!

cadecairos: > Oh I just split the queries into two chainable functions. The prefetch query is the same though. I needed one without the prefetch because it ran too many queries for a query in which I just needed moderated entries that were filtered.

can you unpack this a bit? it's hard to follow

gideonthomas: We did two things in one function - we prefetched a bunch of stuff for entries AND we filtered based on moderation state. Since these are two completely unrelated tasks, I split them into two functions. Everywhere we expected both tasks to be run now instead call both these functions to run those tasks. The tasks are still run as a single query because of the way queryset chaining works. But, I had to get entries that were filtered based on moderation state but I didn't want it to prefetch all those related fields (tags, issues, etc.) because I didn't need them and needed to create a separate prefetch clause.

What @patjouk is referring to was that I think I made a rebase mistake and overrode her changes with the old code. I reverted that so that my change now is to only split one function into two functions without changing any of the queries.

cadecairos: Got it, thanks!

alanmoo:
Where does this stand? Requests to https://api.mozillapulse.org/api/pulse/profiles/?profile_type=fellow are taking way too long to load from the foundation site and I believe this PR will enable us to improve that with lighter requests. Does that sound right?

cadecairos:
I'll review this once more tomorrow, if @gideonthomas says it's still ready for that.

gideonthomas:
yep it's ready!

cadecairos:
testing it now!

gideonthomas:
@cadecairos I'm making one small change to add a v2 to the profiles list view. Shouldn't affect the instructions I have above, but just a heads up that there might be one more thing to test

pulseapi/settings.py
```diff
@@ -201,6 +201,7 @@ def show_toolbar(request):
 # and <version value> is the value of the version that will be used in URLs
 API_VERSION_LIST = [
     ('version_1', 'v1',),
+    ('version_2', 'v2',),
 ]
 DEFAULT_VERSION = 'version_1'
```
cadecairos: I presume we'll default to v2 when the front end catches up?

gideonthomas: I don't have a pref, up to you and @alanmoo

alanmoo: Once all our properties are using versioning in the routes they call, I'd say we can update the default to always reflect the latest version.

gideonthomas:
@cadecairos I might need more info here coz I can't seem to reproduce the error you're running into. Maybe we can debug over vidyo?

gideonthomas:
@cadecairos here are a few things to test:

1. `/api/pulse/profiles/:id/` - returns profile info for that id AND all the entries (published and created) associated with that profile
2. `/api/pulse/v2/profiles/:id/` - returns only profile info without any related entry information
3. `/api/pulse/profiles/:id/entries/?created` - returns the created entries for that profile
4. `/api/pulse/profiles/:id/entries/?published` - returns the published entries for that profile
5. `/api/pulse/profiles/:id/entries/?favorited` - returns the favorited entries for that profile
6. `/api/pulse/profiles/:id/entries/` - returns the number of entries related to that profile (created/published/favorited)
7. `/api/pulse/profiles/?profile_type=<something>` - returns a list of all the profiles AND their related entries (published and created)
8. `/api/pulse/v2/profiles/?profile_type=<something>` - returns a list of all the profiles WITHOUT their related entries.

xmatthewx:
Wow. This is thorough. Nice work @gideonthomas. 

>  6. /api/pulse/profiles/:id/entries/ - returns the number of entries related to that profile (created/published/favorited)

Does this return a count of all entries, or the entries themselves?

I can see a scenario where we'll want created && published for profile. But not a scenario where we'll want created && published && favs.

pulseapi/tests.py
```diff
@@ -164,6 +164,8 @@ class PulseMemberTestCase(TestCase):
     """
     A test case wrapper for "plain users" without any staff or admin rights
     """
+    maxDiff = None
```
gideonthomas: I made this small change to make error logging during tests a bit more informative

gideonthomas:
`/api/pulse/profiles/:id/entries/` only returns the count of all the entries.

We can get the created and published entries for a profile by hitting up `/api/pulse/profiles/:id/entries/?created&published`

alanmoo:
> /api/pulse/profiles/:id/entries/ only returns the count of all the entries.

> We can get the created and published entries for a profile by hitting up /api/pulse/profiles/:id/entries/?created&published

Hold up: aren't we using that first endpoint in pulse currently? That seems like a change that should be versioned.

gideonthomas:
@cadecairos when you get a chance tomorrow, would you be able to review my last commit? It does two things:
1. Fixes the url redirect issue you were running into. Now, `/api/pulse/profiles/:id/entries` will correctly redirect to `/api/pulse/profiles/:id/entries/`.
2. I noticed that the profile listing route I added (e.g. `/api/pulse/v2/profiles?profile_type=fellow`) was still really slow. It was still fetching records individually and so was running a ridiculous number of queries (for `profile_type=fellow` with prod. data it ran 487 queries). In this commit I added prefetching to get it to run in O(1) time with 7 queries.

cadecairos:
@gideonthomas R+ to that last Commit

PR #322, opened 2018-03-06
Adding missing migration + Review app fix

patjouk:
- Fix a missing migration,
- Review app debug is now at true


cadecairos:
@patjouk I forgot that we identified the INTERNAL_IP setting as needing an update before the debug toolbar would work on review apps. I'll make that a requested change and we can discuss how to implement that.

pulseapi/settings.py
```diff
@@ -51,10 +51,19 @@
...
+# Create a simple function to show Django Debug Toolbar on Review App
+def show_toolbar(request):
+    return request.user.is_staff
+
+
 # Adding support for Heroku review app
 if env('HEROKU_APP_NAME'):
     herokuReviewAppHost = env('HEROKU_APP_NAME') + '.herokuapp.com'
     ALLOWED_HOSTS.append(herokuReviewAppHost)
+    DEBUG_TOOLBAR_CONFIG = {
```
cadecairos: does this override https://github.com/mozilla/network-pulse-api/blob/2616bdabd3f742f1625e436962daad93dcabbf58/pulseapi/settings.py#L113-L118?

PR #326, opened 2018-03-13
Add support for profile response ordering

alanmoo:
And some documentation

### DIFF
--- a/pulseapi/profiles/views.py
+++ b/pulseapi/profiles/views.py
@@ -131,10 +131,18 @@ class ProfileCustomFilter(filters.FilterSet):
 
 
 class UserProfileListAPIView(ListAPIView):
+    """
+      Query Params:
+      profile_type=
+      program_type=
+      program_year=
+      ordering={property to sort by (negative to reverse)}
+    """
     serializer_class = UserProfilePublicSerializer
 
     filter_backends = (
         filters.DjangoFilterBackend,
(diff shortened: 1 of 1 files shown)


mmmavis:
https://pulse-api.mofostaging.net/api/pulse/profiles/?profile_type=fellow&ordering=name is giving me status `500`. Did I pass the param in the right way? @gideonthomas @alanmoo

gideonthomas:
I don't think drf on its own is equipped to deal with fields like `name` that aren't actual fields on the profiles model. We might have to subclass OrderingFilter and write custom logic to allow ordering by name. Instead, I suggest just doing this in the serializer directly and always order by name instead of allowing you to specify how to order them by via url.

alanmoo:
`custom_name` does work, and a random sampling of our fellows' profiles on production shows they seem to all have custom names. So let's roll with that for now and fix any that are missing it manually.

I'm not sure if the real solution to this is to write a new filter or to straighten out our `custom_name`/`name` situation, but it feels low priority either way. Thoughts @gideonthomas?

gideonthomas:
I'd rather it not 500 at all (so that it doesn't unnecessarily trigger devops paging) so if you're going with that solution, make sure that `ordering_fields` is specified to only allow `custom_name`. That would also mean that we should document that it only allows ordering by `custom_name`.

PR #329, opened 2018-03-15
Check for the proper user permissions on EmailUsers when moderating

cadecairos:
Seems we're checking the wrong permissions when determining if an authenticated user can moderate.

This should fix the issue where email users need super user perms to moderate, but in order for it to work, we need to double check that the `moderators` group has the `entries.change_entry` permission, and all the EmailUsers we want to be able to moderate should be in it.


cadecairos:
This needs a bit more work. It seems I've broken some bad expectations in the tests. they also only seem to test the "staff" group, so I'm going to see what testing the "moderator" group will look like.

pulseapi/entries/views.py
```diff
@@ -64,7 +64,7 @@ def toggle_featured(request, entryid, **kwargs):
     """
     user = request.user
 
-    if user.has_perm('entries.can_change_entry'):
+    if user.has_perm('entries.change_entry'):
```
cadecairos: `can_change_entry` was programmatically created in the tests, but the actual permission for changing entries has the codename `change_entry`

PR #330, opened 2018-03-15
Make name filtering not 500

alanmoo:
Follow up to #326

### DIFF
--- a/README.md
+++ b/README.md
@@ -529,8 +529,7 @@ While for local development we provide a `sample.env` that you can use as defaul
 
 ### Review App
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -528,7 +529,7 @@ While for local development we provide a `sample.env` that you can use as defaul
 
 ### Review App
 
-Opening a PR will automatically create a Review App in the `network-pulse-api` pipeline. It's not possible to 
+Opening a PR will automatically create a Review App in the `network-pulse-api` pipeline. It's not possible to
```
gideonthomas: lol can you fix this? I think this has changed in multiple PRs coz the next line is on its own line instead of being a part of this line.

PR #333, opened 2018-03-20
Make lists of entries automatically return in reverse order

### LINKED ISSUE #324: Enable or default to reverse order on entries from profile endpoint
Profiles currently display related entries in chronological order. Reverse chronological would be better, especially for profiles like [Mozilla's](https://www.mozillapulse.org/profile/209) that contain a dozen or more items.

The entries endpoint allows ordering with `?ordering=-created`. But they `published_entries` and `created_entries` cannot be ordered on `/profiles/<profile_id>` 

Again, default to reverse is fine. I don't see a big need for custom ordering, but that's fine too. 

Unblocks https://github.com/mozilla/network-pulse/issues/934

cc @alanmoo to triage and assign.

alanmoo:
Especially useful for profiles.
Closes #324 
I believe it automatically fixes https://github.com/mozilla/network-pulse/issues/934 too.

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -145,7 +145,7 @@ class UserProfilePublicSerializer(UserProfileSerializer):
         user = instance.user
 
         return EntrySerializer(
-            user.entries.public(),
+            user.entries.public().order_by('-id'),
             context=self.context,
             many=True,
         ).data if user else []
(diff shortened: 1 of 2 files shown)


xmatthewx:
Yes, defaulting to reverse order will resolve the frontend issue as well.

pulseapi/entries/models.py
```diff
@@ -174,7 +174,7 @@ class Meta:
         Make plural not be wrong
         """
         verbose_name_plural = "entries"
-        ordering = ['id']
+        ordering = ['-id']
```
gideonthomas: let's not do this at the model level coz it forces `order by` clauses on querysets that don't need it

gideonthomas: consequently, add the order_by clause for `published_entries` as well

alanmoo: It's already forcing the clause, I'm just changing the default. Besides, the `published_entries` aren't fetched the same way as `created_entries` so I can't seem to simply stick on `ordered_by` anywhere obvious.

gideonthomas: ok, if we're sure we want reverse chronological ordering for every type of fetch on a list of entries then this is fine. Otherwise, `published_entries` are fetched here: https://github.com/mozilla/network-pulse-api/pull/333/files#diff-9caaebc8dbfd9b041d5714a6e437bd47R148, you just need to chain the `order_by` query there.

PR #347, opened 2018-04-18
Generate a configurable amount of fellows in varying years and programs

cadecairos:
This PR adds some code to the `load_fake_data` command that will iterate over program types, and program years to generate a random set of fake Mozilla Fellows.

By default it will generate 3 fellows per program per year, but you can tweak that number with the `--fellowsCount` flag

### DIFF
--- a/pulseapi/utility/management/commands/load_fake_data.py
+++ b/pulseapi/utility/management/commands/load_fake_data.py
@@ -36,7 +36,7 @@ class Command(BaseCommand):
         )
 
         parser.add_argument(
-            '--fellowsCount',
+            '--fellows-count',
             action='store',
             type=int,
             default=3,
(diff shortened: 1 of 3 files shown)


pulseapi/utility/management/commands/load_fake_data.py
```diff
@@ -35,6 +35,15 @@ def add_arguments(self, parser):
             help='A seed value to pass to Faker before generating data'
         )
 
+        parser.add_argument(
+            '--fellowsCount',
```
patjouk: Can you change to "fellows-count"?

cadecairos: yup. Can you tell I'm used to PowerShell (Although, it would have been `-FellowsCount` if a Powershell utility)

PR #348, opened 2018-04-19
Replace creators with profiles

### LINKED ISSUE #319: Discuss admin of creators and profiles
Let's discuss any follow up work needed for how admins can associate creators and profiles. 

I have a few comments below on @gideonthomas's [PR notes](https://github.com/mozilla/network-pulse-api/pull/318#issue-172372422). In sum, 
  - I think the PR does the job of unblocking us on fellow profiles now. So that's good for now. I think we've  created new risks of user error. 
  - I wonder whether we need to step back and rethink the root of this. For example, if all existing and new Creators had a profile, empty and inactive, this might all be less complex. An active/inactive profile state seems easier to manage than shifting associations between models. 

I think we should push the PR after testing, and address the unintended consequences as a follow up with more time for all of us to think through things.  

Let's meet to discuss over video. A few rough notes for conversation (no need to reply): 
 - New Profiles can now be associated with existing Creators. This is great, and unblocks us. 
 - New and existing Profiles can be associated with Creators that already have a Profile. This is prone to user error and will create orphaned data an admin cannot correct. 
    - Admins should edit existing profiles rather than creating a new one. We risk duplicate profiles and unintentionally orphaned profiles still visible on the frontend. 
    - Admins could orphan a Creator. These creators will remain associated with entries and will be offered in  autocomplete on the new entry form.  

Lastly, I can't think of a use case for this:

> Update an existing profile and make sure that you can change its creator to the option that automatically creates a new creator.

gideonthomas:
**Pending**: Updating the README

This PR effectively removes the concept of "creators" for entries. Because of that unnecessary abstraction, we had extra models - `Creator` and `OrderedCreatorRecord` to maintain a relationship between profiles and entries. Those models have been replaced with a single `EntryCreator` model to handle the many-to-many relationship with entries.

#### Independent tests

1. Load fake data (or prod data) and then run `manage.py migrate`. Make sure it runs without a hitch. Check your local db to make sure that the `creators_creator` and `creators_orderedcreatorrecord` tables no longer exist and that the `creators_entrycreator` table is added and has data.
2. Run `manage.py test` to make sure that tests are running correctly without any random failures.
3. Go to the admin and open a random profile. Make sure you don't see a "Creator" field in the profile.

For the tasks below, it will be useful if you have the front-end setup to create entries, unless you're using production data.

#### Existing API route tests

1. `GET /creators/?name=<search string>` - Login and open the front-end form to create a new entry. Make sure when entering creators, you see autocomplete suggestions after entering 3 letters.
2. `POST /entries/` - Create an entry with creators (use a mix of autocompleted creators and typed in creators). Hit submit and make sure it has submitted without any errors.
3. `GET /entries/<id>/` - Confirm that you see the entry on the front-end with the related creators. Note the entry ID.
4. `POST /entries/bookmarks/` - Try favoriting an entry and make sure there are no console errors.
5. `GET /entries/bookmarks/` - Go to your profile and make sure you can see your bookmarked entries.
6. `GET /profiles/<id>/` - Go to someone else's profile and make sure you see all the relevant profile information.

#### Tests for new API routes / Routes not used by the front-end

These tests don't require the front-end so you would directly go to these routes in your browser and inspect the result. You will also use the entry (whose id you noted) you created above here as a test case. Make sure you're logged into the API.

1. `GET /v2/creators/?name=<search string>` - Make sure this returns an error payload since we're deprecating this route.
2. `GET /v2/entries/<id>/` - Go to the `id` of the entry you created. Make sure the result has all the relevant entry data and that the `related_creators` property exists as an array of objects where each object contains a `name` and a `profile_id`. Assert that these creator objects do not have the `creator_id` property.
3. `GET /v2/entries/bookmarks/` - This should return an array of entry objects (each one being the same format as 2. above) that you bookmarked.
4. `GET /v2/entries/` - Should return a paginated array of entry objects (each one being the same format as 2. above).
5. `GET /v2/profiles/<choose a creator's profile id from the entry above>/entries/?created` - Should return an object with a `created` property whose value is an array of entry objects. Each entry object should NOT be the same format as 2. above, but instead should only have 4 basic details about the entry + a `related_creators` property whose value is an array of objects each with a `name` and `profile_id`. Verify that this route only runs 6-7 SQL queries.
6. `GET /v2/profiles/?name=<search string>` - Type in a couple of letters (or just 1) in a creator's name as the search string and make sure that an array of profile objects is returned where each object contains all the profile information including `thumbnail`, `github`, etc.
7. `GET /v2/profiles/?name=<search string>&basic` - Use the same search string as above and tack in `&basic` in the URL. You should get the same result as above except each profile object only has the `id` and `name` properties.
8. `GET /v2/profiles/?name=<search string>&is_active=true` - Go to the admin and modify a few profiles by setting the "is active" flag on them. Hit the route mentioned with `&is_active=true` tacked onto the URL and make sure that the result is an array of profile objects that you modified in the admin, where each object contains all the profile information including `thumbnail`, `github`, etc.

Fix #319


xmatthewx:
@gideonthomas - when you're ready, is it useful to pull together a group of people to help test? there's a lot in here!

gideonthomas:
I think if there was a front-end for this it might be useful to get a group to test. Otherwise, if you want to split up the test script above between people, it might help @cadecairos.

gideonthomas:
@xmatthewx, just an fyi, before we land this, we should flip the `is_active` boolean for every profile on staging and production so that we can distinguish them from creators profiles who weren't created by a person (i.e. auto-generated when a creator was mentioned in an entry). Should I do this at the database-level with an sql query.

xmatthewx:
@gideonthomas – good catch. i don't think that boolean has been used, or maybe just a few times  on fellows.

PR #351, opened 2018-04-23
Change to optimized entry selector in tag admin

### LINKED ISSUE #342: Add Pagination to change Tag view
Pulse admin allows us to view a tag, see list of associated entries, and a new entry association. 

For some tags with a long list of associated entries, like mozsprint (id 400), the page barely loads, sometimes fails. see: https://api.mozillapulse.org/admin/tags/tag/400/change/

Can we paginate this view? 

@zee-moz is actively using this to triage submission around global sprint. 

Note: you can edit an entry to select associated tags, but the multi-select box is very fiddly and dangerous.

gideonthomas:
Fix #342 

Previously, you could not even load a tag page in the admin to change it because it took too long to render the entries it is associated to.

I couldn't find a way to add pagination for inline model admins. So, I just changed it to a multi selector for now. 

This PR also adds a search bar to the Tag and Entry admin coz it was impossible to find tags.

#### Things to test in the admin
1. Create a new tag and associate a few entries to it and save the tag. Make sure that the tag was successfully associated with the entries.
2. Edit an existing tag to change (add/remove) the entries associated with it. Make sure the changes persist.
3. You can search for tags when you look at the tag list in the admin.
4. You can search for entries using tags or the title when you look at the entry list in the admin

### DIFF
--- a/pulseapi/entries/admin.py
+++ b/pulseapi/entries/admin.py
@@ -70,6 +70,11 @@ class EntryAdmin(admin.ModelAdmin):
         'featured'
     )
 
+    search_fields = (
+        'title',
+        'tags__name',
+    )
+
     def bookmark_count(self, instance):
         """
         Show the total number of bookmarks for this Entry


xmatthewx:
I'm curious what @acabunoc thinks about this. 

Entry edit view already has a multi-select box for tags, which are a fiddly pain and can dangerously lose data on user interaction. I'm unsure if we should change to multi-select box on the Tag edit view. While loading sometimes fails, the interface is way better.  

I'm fine either way, just want to make sure we're making the right choice for our gold star user @acabunoc.

acabunoc:
@gideonthomas are you in today? find me in the kitchen and demo this for me? 😃

gideonthomas:
@xmatthewx, confirmed with @acabunoc that this solution is fine. I'm not a huge fan of the UX either but I think as a temporary solution this might suffice.

acabunoc:
I LOVE IT 😍 👏 💯 ✨

xmatthewx:
thanks @gideonthomas for the fix and the diligence!

PR #365, opened 2018-05-08
Introducing pipenv

### LINKED ISSUE #203: Update README.md with instruction to use python 3.x for this project
We should probably add the following to README.md (copied from the network-api README)
---

#### Important note for systems with python *and* python3

In order to make sure your virtual environment will be using python 3.x you will have to explicitly tell the system it should use point to `python3` whenever it invokes python:

```
$ virtualenv -p python3 venv
```

### LINKED ISSUE #210: update package.json to be platform-agnostic
package.json currently supports *n*x flavours, but not windows. We will want to update the npm scripts in a way that lets any dev run the convenience scripts.

### LINKED ISSUE #360: pyflakes 1.4.0 is incompatible with flake8 3.4.1
Got this error from `pip install -r requirements.txt`. We should probably update our `requirements.txt`

screen shot 2018-05-01 at 2 49 33 pm

### LINKED ISSUE #364: Moving to pipenv
Following the work done [on the Foundation site](https://github.com/mozilla/foundation.mozilla.org/pull/1320), let's move to pipenv.

patjouk:
Close #364, close #360, close #203, close #210
Todos:
- [x] Update and unpin dependencies
- [x] Comment on what can't be easily updated/unpinned
- [x] Fix minor lint issues
- [x] Remove req.txt
- [x] Replace pip in CI
- [x] Replace pip in deploy
- [x] Write invoke tasks (not compatible with windows for now 😿 cf https://github.com/mozilla/foundation.mozilla.org/issues/1529)
- [x] Update the doc
- [x] Add fix for invoke + rewrite docs

~Fix bare except issues~ moved to a [separate issue](https://github.com/mozilla/network-pulse-api/issues/368)


Pomax:
as far as npm tasks go, argument forwarding is possible using `--` so we could have an npm script `"manage": "pipenv run python manage.py" which can then be invoked as `npm run manage -- runserver`. `npm run manage -- makemigrations` etc.

patjouk:
I've fixed Invoke tasks so that `inv runserver` output is acting normally again.

patjouk:
Testing it locally with pulse front, I ran into an issue around CORS: since this issue is not on master, I started poking around and found this:

screen shot 2018-05-25 at 4 16 49 pm

Something is adding wayyyy too many `\` in that regex env var 😆 
The issue is in one of pipenv dependencies, python-dotenv, and [someone already reported it](https://github.com/theskumar/python-dotenv/issues/112) 11 days ago. The workaround is to temporarily deactivate the autoload of the `.env` file by doing: `PIPENV_DONT_LOAD_ENV=1 pipenv run ./manage.py runserver`. I'll change the invoke task.

cadecairos:
> Does the .env file still get used eventhough pipenv itself has been set to not load it?

We can confirm it, but I'm pretty sure settings.py will try to load it in.

patjouk:
It will work as usual (I tried it before doing the commit).

PR #374, opened 2018-05-22
Add entry count to profile routes

gideonthomas:
Adds an entry count for published, created, and favorited entries for profiles. Also added prefetching to the /myprofile route to reduce from 21 queries to 9.

To test:
1. Test the `api/pulse/profiles/<id>` routes to make sure that they include the `entry_count` object
2. Test the `api/pulse/myprofile` route to make sure that it also includes the entry count and is running less than 10 queries.

@mmmavis would you be able to test the routes you need for the profile view to make sure they have all the information you need?


gideonthomas:
@mmmavis nice catch! I think that was because I wasn't filtering out entries that weren't approved. This should be fixed now...would you be able to verify?

gideonthomas:
Thanks @mmmavis! @Pomax would you be able to review this pr?

pulseapi/profiles/serializers.py
```diff
@@ -100,6 +100,19 @@ class UserProfileSerializer(serializers.ModelSerializer):
...
+    entry_count = serializers.SerializerMethodField()
+
+    def get_entry_count(self, obj):
+        entry_queryset = Entry.objects.public()
+
+        return {
+            'created': obj.related_entry_creators.filter(entry__in=entry_queryset).count(),
+            'published': entry_queryset.filter(
+                published_by=obj.related_user
+            ).count() if getattr(obj, 'related_user', None) else 0,
```
Pomax: Just for logic's sake, `if getattr(obj, 'related_user', False) else 0` probably makes more sense?

PR #376, opened 2018-05-31
Interactive mode for factories

### LINKED ISSUE #344: Interactive mode for the fake data factories
`load-fake-data` create a fixed number of fake creators, entries, fellows, etc. We might want to create a different mode that would let people enter the number of elements they want. Of course, we will keep an automatic mode for the review apps.

We could use [click-cli](http://click.pocoo.org/5/) and [django click](https://pypi.python.org/pypi/django-click/2.0.0).

@gideonthomas

patjouk:
close #344 

Rewrote the `load_fake_data` management command: it now takes flags like `--entries-count` to change how many fellows we want.
Removed the `generate-fellows` management command for now (need discussion).

Available flags:
- `-f`, `--fellows-count`: The number of fellows to generate per program type, per year. Defaults to 3
- `-u`, `--users-count`: The number of users to generate per possible variations. Default: 1, variations: 80
- `-e`, `--entries-count`: The number of entries to generate per possible variations. Default: 20, variations: 16
- `-t`, `--tags-count`: The number of tags to generate. Default: 6


patjouk:
@gideonthomas: I added flags to the `load_fake_data` commands: it will run the script as usual but changes the default values to the one you give. I can stop here or add separate commands, like the [`generate-fellows` commands](https://github.com/mozilla/network-pulse-api/blob/master/pulseapi/utility/management/commands/generate_fellows.py). Do you need commands that will generate only entries, or only users? If yes, which ones in particular?

patjouk:
I got this error too from time to time: I thought that switching the group param's faker provider [from color to company](https://github.com/mozilla/network-pulse-api/pull/376/files#diff-d1fa3369df3e78deb2253172444628a7R26) would do the trick 😞
I'll try to reproduce it and find which factories need adjustment.

patjouk:
Found 2 bugs:
- `django.db.utils.IntegrityError: UNIQUE constraint failed: creators_entrycreator.entry_id, creators_entrycreator.profile_id` happens when the script tries to link a creator to the same entry again. If that happens, it now skips that entry and continues the creators/entries for-loop.
- `django.db.utils.IntegrityError: UNIQUE constraint failed: users_emailuser.email`: happens on MozillaUsers creation. Since we limit the domain to `@mozillafoundation.org`, the script has less room to create a non-existing email address. Now, if that happens, it stops creating new users for the run and moves on to the rest of the script.

The first error was happening A LOT (how did I miss that!), the second one was more "bad luck". I've run the script more than 10 times (without deleting between runs) and have now 624 `@mozillafoundation` users on 1813 in total.

patjouk:
After discussing it with @cade, I will add a random number to the `group` field in the profile factories to increase the randomness and prevent "unique constraint" error.

cade:
Increased randomness, eh?

Is it crazy how saying sentences backwards creates backwards sentences saying how crazy it is?

Here's a turtle, covered in moss:

patjouk:
@cade Oh gosh I'm sorry >.< They told me I would ping the wrong cade at some point! I'll be more careful :)

cadecairos:
@patjouk I think once we add some defensive measures to prevent hitting that unique constraint error, this is good to go.

PR #379, opened 2018-06-19
Add catch-up task and fix setup task

patjouk:
Ref mozilla/foundation.mozilla.org#1575

Running `inv setup` on an already configured project would silently replace your `.env` by the default one. It's now fixed: the script will stop if a `.env` is already there and proposes two solutions to the user.
I added a `inv catch-up` command that will only run `pipenv install` and apply migrations.

If this solution is okay, I will apply the same changes to the `tasks.py` of the foundation site.


Pomax:
that should have been "sweet!" but you can't edit a review comment. For reasons I don't understand.

PR #384, opened 2018-07-03
Order profile entries

### LINKED ISSUE #370: Allow custom ordering of profile entries
For the `/profiles/<id>/entries/` route, let's allow specifying an ordering flag for each of the three types of entries, something like `?favorited_ordering=`, `?created_ordering=`, and `?published_ordering=`.

gideonthomas:
Fix #370 

This PR also fixes a regression in performance for getting profile entries where we were running over 55 SQL queries before because we weren't prefetching bookmarks.

To test:
1. Run migrations, and populate data (either from prod to test performance or using fake data)
2. Hit up `/api/pulse/v2/profiles/<id>/entries/?created&created_ordering=-created` and make sure that the entries returned are in reverse chronological order (you can rely on the `id` to verify that)
3. Repeat the above with `?published&published_ordering=-id` and `?favorited&favorited_ordering=title` (for this last one just make sure that the entries are ordered alphabetically by their title).

### DIFF
--- a/pulseapi/profiles/serializers.py
+++ b/pulseapi/profiles/serializers.py
@@ -238,7 +238,8 @@ class UserProfileEntriesSerializer(serializers.Serializer):
             }
 
         entry_queryset = Entry.objects.prefetch_related(
-            'related_entry_creators__profile__related_user'
+            'related_entry_creators__profile__related_user',
+            'bookmarked_by__profile__related_user',
         )
 
         if include_created:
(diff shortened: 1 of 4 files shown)


pulseapi/profiles/serializers.py
```diff
@@ -215,6 +215,18 @@ class UserProfileEntriesSerializer(serializers.Serializer):
     returning the number of entries (created, published, and favorited) associated
     with the profile.
     """
+    @staticmethod
+    def get_ordered_queryset(queryset, ordering_param=None, prefix=False):
+        if not ordering_param:
```
Pomax: should this be `if ordering_param is None`? Can it ever something falsey that isn't None?

gideonthomas: not really and if it does somehow happen to be falsey (like `''`) it should still only return the queryset as is instead of throwing an error and crashing the server when ordering by a falsey value.

PR #385, opened 2018-07-04
Python runtime update

patjouk:


cadecairos:
Not sure what's up with appveyor, but as per a comment on another issue with the same error, you could try specifying a 64 bit environment for the build - it looks like it's defaulting to a 32 bit build box.

PR #386, opened 2018-07-04
update dependencies + django

patjouk:
ref: MozillaFoundation/mofo-devops-private#146


patjouk:
The issue was around `python -m pipenv run manage.py migrate`: I forgot the `python` after the `pipenv run ` >.<

PR #391, opened 2018-07-30
Add is_active flag to serialized entry creators

### LINKED ISSUE #389: Add is_active to entry related creators
We need to add the flag to the serialized data to make sure the front-end doesn't link to it.

gideonthomas:
Fix #389 

Since this change is additive, it doesn't need to be versioned.

To test:
1. Create an entry and for its creators, autocomplete some profiles and type in the rest (so they aren't existing profiles).
2. Hit up `/api/pulse/v2/profiles/<your profile id>/entries/?published` and check that the `related_creators` for your entry has the correct boolean for `is_active`.

either @Pomax or @alanmoo can review...whoever has the bandwidth to do so.

### DIFF
--- a/README.md
+++ b/README.md
@@ -202,6 +202,7 @@ Each `related_creator` object will have the following schema:
 {
   name: <string: name of the creator profile>
   profile_id: <integer: id of the creator profile>
+  is_active: <boolean: flag indicating whether this creator profile is attached to a user>
 }
 ```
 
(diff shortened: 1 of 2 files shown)


alanmoo:
I'm not quite sure what's happening but when I visit https://network-pulse-api-stagi-pr-391.herokuapp.com/admin/profiles/userprofile/ to check the existing profiles I get a Heroku error. I did see this page once on this review app, but it just...stopped working even though I never went into an actual profile. I don't think this is caused by the PR (the code looks fine to me, but I still want to test it).

gideonthomas:
@alanmoo I didn't think review apps worked for pulse api, but maybe I'm wrong? I remember having to use local dev to test stuff on the api.

PR #393, opened 2018-07-31
Filter entries by the presence of help types

### LINKED ISSUE #392: Allow filtering out entries without any help types
Add a filter to the entry list to filter entries based on whether they have help types associated with them or not.

gideonthomas:
Fix #392

To test (if needed although the automated test pretty much covers it):
1. Create an entry with a few help types and another without help types.
2. Hit up `/api/pulse/v2/entries/?has_help_types=True` and make sure the entry with help types is present. Do the same with `False` and make sure the entry without help types is present.
3. r+ my patch :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -238,6 +238,7 @@ The schema for the payload returned is:
 - `?tag=<string>` - Filter entries by a specific tag
 - `?issue=<string>` - Filter entries by an issue area
 - `?help_type=<string>` - Filter entries by a specific help category
+- `?has_help_types=<True or False>` - Filter entries by whether they have help types or not. Note that `True` or `False` is case-sensitive.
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -238,6 +238,7 @@ The schema for the payload returned is:
 - `?tag=<string>` - Filter entries by a specific tag
 - `?issue=<string>` - Filter entries by an issue area
 - `?help_type=<string>` - Filter entries by a specific help category
+- `?has_help_types=<True or False>` - Filter entries by whether they have help types or not. Note that `True` or `False` is case-sensitive.
```
alanmoo: Can you also document this in `entries/views.py` so it shows up at the top of `/api/pulse/entries`?

gideonthomas: done

PR #395, opened 2018-08-02
profile categories endpoint

### LINKED ISSUE #394: Expose profile categories
We need an endpoint that gives us options we can use when filtering profiles. It should expose Profile types, Program types, and Program years.
Something like:
```
{
'profile_types':['fellow','staff','etc'],
'program_types: ['Media','Tech Policy', 'etc'],
'program_years:['2015','2016', 'etc']
}
```

cadecairos:
This PR creates an endpoint that lists all the possible profile and program types, as well as program years. Includes a test.

The endpoint resides at: `/api/pulse/profiles/categories`

Fixes #394

### DIFF
--- a/pulseapi/profiles/test_views.py
+++ b/pulseapi/profiles/test_views.py
@@ -2,6 +2,7 @@ import json
 from urllib.parse import urlencode
 from django.core.urlresolvers import reverse
 from django.conf import settings
+from collections import namedtuple
 
 from .models import UserProfile, ProfileType, ProgramType, ProgramYear
 
@@ -23,6 +24,11 @@ from pulseapi.profiles.serializers import (
     UserProfileBasicSerializer,
 )
 from pulseapi.profiles.factory import UserBookmarksFactory, ExtendedUserProfileFactory
+
+
+# Converts the JSON String to a named tuple - The name of the tuple (X) is unimportant
+def json_to_obj(data):
+    return json.loads(data, object_hook=lambda data: namedtuple('X', data.keys())(*data.values()))
 
 
 class TestProfileView(PulseMemberTestCase):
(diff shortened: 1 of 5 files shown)


cadecairos:
Endpoint in action on the Review App: https://network-pulse-api-stagi-pr-395.herokuapp.com/api/pulse/profiles/categories/

pulseapi/profiles/views.py
```diff
@@ -252,3 +257,16 @@ def get_serializer_context(self):
         return {
             'user': self.request.user
         }
+
+
+class UserProfileCategoriesView(APIView):
+    def get(self, *args, **kwargs):
```
gideonthomas: fyi: the reason you probably were getting errors for this signature is because of the inheritance from `APIView`. If your signature is `def get(self, request, **kwargs):` I think it should work.

pulseapi/profiles/test_views.py
```diff
@@ -433,3 +439,29 @@ def test_invalid_profile_list_filtering_argument(self):
         response = self.client.get(url)
         entriesjson = json.loads(str(response.content, 'utf-8'))
         self.assertEqual(len(entriesjson), 0)
+
+    def test_profile_categories(self):
+        profile_categories_url = reverse('categories_view')
+        response = self.client.get(profile_categories_url)
+        categories = json_to_obj(str(response.content, 'utf-8'))
+        self.assertListEqual(categories.profile_types, [
```
gideonthomas: let's not hardcode these. Might as well just use the same query you do in the view method.

pulseapi/profiles/test_views.py
```diff
@@ -25,6 +26,11 @@
 from pulseapi.profiles.factory import UserBookmarksFactory, ExtendedUserProfileFactory
 
 
+# Converts the JSON String to a named tuple - The name of the tuple (X) is unimportant
+def json_to_obj(data):
+    return json.loads(data, object_hook=lambda data: namedtuple('X', data.keys())(*data.values()))
```
gideonthomas: not sure I understand this. Why do we want a named tuple?

cadecairos: honestly, not sure we do? I looked for a method of converting JSON to a python object, this is what I found. let me try with tuple.

cadecairos:
@gideonthomas I've updated the PR, have a look!

cadecairos:
@gideonthomas Docs added!

gideonthomas:
Thanks @cadecairos!

PR #406, opened 2018-09-05
 added is_active flag to profile serializer

### LINKED ISSUE #404: Add is_active flag to profile serializer
Let's return the `is_active` flag in all the profile endpoints for `v2` of the API.

Use Case: https://github.com/mozilla/network-pulse/issues/980#issuecomment-413727349

ali-alsabbah:
Fix #404


pulseapi/profiles/serializers.py
```diff
@@ -96,6 +96,11 @@ class UserProfileSerializer(serializers.ModelSerializer):
         allow_blank=True,
     )
 
+    is_active = serializers.BooleanField(
+        required=False,
```
gideonthomas: let's make this read_only as well.

gideonthomas: you also need to remove this from the `exclude` option in the `Meta` class

pulseapi/profiles/serializers.py
```diff
@@ -162,6 +167,7 @@ def to_representation(self, obj):
         return {
             'id': obj.id,
             'name': obj.name
```
gideonthomas: always end every pair in a dict with a comma

pulseapi/profiles/serializers.py
ali-alsabbah: done

pulseapi/profiles/serializers.py
ali-alsabbah: done

pulseapi/profiles/serializers.py
```diff
@@ -96,6 +96,11 @@ class UserProfileSerializer(serializers.ModelSerializer):
         allow_blank=True,
     )
 
+    is_active = serializers.BooleanField(
+        required=False,
+        allow_blank=True,
```
gideonthomas: remove this since it's not a valid option for `BooleanField`

gideonthomas: yeah probably...I don't think it needs the `required` option since it will always be there.

ali-alsabbah: done

pulseapi/profiles/serializers.py
```diff
@@ -96,6 +96,9 @@ class UserProfileSerializer(serializers.ModelSerializer):
         allow_blank=True,
     )
 
+    is_active = serializers.BooleanField(
+    )
```
gideonthomas: this shouldn't be on a new line if it isn't taking any parameters.

ali-alsabbah: fixed

gideonthomas:
Thanks @ali-alsabbah!

PR #408, opened 2018-09-10
add ?ids=... filtering for the api/pulse/profiles/ end point

Pomax:
required by the foudation.moz work on showing user profiles.

### DIFF
--- a/pulseapi/profiles/models.py
+++ b/pulseapi/profiles/models.py
@@ -78,6 +78,12 @@ class UserProfileQuerySet(models.query.QuerySet):
     A queryset for profiles with convenience queries
     """
 
+    def limit(self, value):
+        """
+        Return only the last [limit] number of entries created.
+        """
+        return self.order_by('-id')[:value]
+
     def active(self):
         """
         Return all profiles that have the is_active flag set to True
--- a/pulseapi/profiles/models.py
+++ b/pulseapi/profiles/models.py
@@ -78,11 +78,11 @@ class UserProfileQuerySet(models.query.QuerySet):
     A queryset for profiles with convenience queries
     """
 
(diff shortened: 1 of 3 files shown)


Pomax:
@gideonthomas it would be nice if we could also say "and don't give me any entries/faves/etc information" so that the profiles stay small. Would that be a lot of work?

gideonthomas:
If you use `v2` of the API, it should already do that, e.g. https://api.mozillapulse.org/api/pulse/v2/profiles/?profile_type=plain

pulseapi/profiles/views.py
```diff
@@ -185,29 +194,34 @@ def qs(self):
...
-            return empty_set
+            return qs
 
         fields = ProfileCustomFilter.get_fields()
         for key in fields:
             if key in queries:
-                return super(ProfileCustomFilter, self).qs
+                qs = super(ProfileCustomFilter, self).qs
+
+        if 'limit' in queries:
```
gideonthomas: can you make this a filter property? So it would be:

```python
limit = django_filters.NumberFilter(method='filter_limit')

def filter_limit(self, queryset, name, value):
    return queryset.limit(value)
```

pulseapi/profiles/models.py
```diff
@@ -78,6 +78,12 @@ class UserProfileQuerySet(models.query.QuerySet):
     A queryset for profiles with convenience queries
     """
 
+    def limit(self, value):
+        """
+        Return only the last [limit] number of entries created.
+        """
+        return self.order_by('-id')[:value]
```
gideonthomas: why do you need `order_by` here? The profile insertion order doesn't really have any significance.

PR #421, opened 2018-12-11
switch to pipenv sync

patjouk:
close MozillaFoundation/mofo-devops#638


patjouk:
After discussion, we'll keep `install -d --deploy` since it checks if the Pipefile is out of sync with its lockfile.

PR #460, opened 2019-02-21
python updates

patjouk:
close #452, close #450, close #448, close #444


patjouk:
linter issues >.< I'll fix that tomorrow

patjouk:
It's a bit more complicated than lint issues and will require investigating what's really going on there. I'm opening a separate issue for now.

PR #464, opened 2019-03-12
Enable profile search

### LINKED ISSUE #453: Enable search of Profiles
Pulse now has 100+ fellows and 2000 profiles overall and 500 that include at least one social link. Useful stuff! Let's make it more useful by enabling search for profiles. 

To keep things simple, we can keep this search separate from project search. We will figure out a simple way for users to toggle between them. 

 - Search fields: `custom_name, affiliation, user_bio, user_bio_long, location`
 - Useful filter field: `profile_type_id`
 - Return by `id desc`

gideonthomas:
Fix #453

Adds v3 for the API so that we can optimize what's returned in the list profile route.

Relevant test:
Run `/api/pulse/v3/profiles/?search=` and search by their name, user bio, affiliation, or location. Make sure that it runs only 3 queries.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,7 @@ To maintain legacy support, if the version is not specified in the url, we defau
 
 ## Supported API Versions
 
+- Version 3 - `/api/pulse/v3/` and `/v3/`
 - Version 2 - `/api/pulse/v2/` and `/v2/`
 - Version 1 - `/api/pulse/v1/` and `/v1/`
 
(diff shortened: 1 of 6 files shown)


alanmoo:
No rush to review this

xmatthewx:
Does this have a limit set? Seems like an empty `search=` yields a long list. 

What's the default sort? It doesn't seem to be `id`. Probably best as `id desc`. 

I can file a new issue, just wanted to make sure I wasn't making incorrect assumptions.

gideonthomas:
Good flags.

> Does this have a limit set? Seems like an empty search= yields a long list.

No limit, the list being long isn't going to be a performance issue. But, I think on the front-end, we shouldn't load all of it at once because loading several images at once will slow down the page. To make it easier for pagination on the front-end, I can also paginate this API -- would you be able to file a new ticket for that?

> What's the default sort? It doesn't seem to be id. Probably best as id desc.

No sorting, we can add id sorting without a performance hit I think. Would you be able to add that to the same ticket I mention above?

PR #469, opened 2019-03-29
Add pagination to v3 of profile list

### LINKED ISSUE #466: Add pagination to profile search
Let's add a limit and pagination to new profile search feature (discussed in https://github.com/mozilla/network-pulse-api/pull/464#issuecomment-476886273). 25 seems like a reasonable number, but feel free to match what we've used for entries. 

While in there, let's set default sorting to `id desc`.

gideonthomas:
Fix #466 

Adds pagination to profile listing for `v3` and above.

To Test:
- Test out v3 of the API for `/profiles/?name=` and make sure that the results are paginated and the order of the profiles returned are in descending order of `id`.
- Test out v2 of the API for `/profiles/?name=` and make sure that the results aren't paginated but the order of the profiles returned are in descending order of `id`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -531,7 +531,7 @@ Depending on the API version specified, the `related_creator` object schema will
 
 ### `GET /api/pulse/profiles/?...` with filter arguments, and optional `format=json`
 
-Returns a list of user profile objects each with the following schema:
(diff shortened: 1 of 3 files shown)


alanmoo:
@cadecairos or @patjouk when one of you has a chance, can you give this a review?

PR #478, opened 2019-04-23
Heroku yml

cadecairos:
This PR sets us up to build and run our own docker containers on Heroku.

Still TODO: determine if it still makes sense to run appveyor tests, if we're containerized.


Pomax:
Closing as older, no longer relevant, work.

Issue #479, opened 2019-04-23
Add Entry "type" field


Pomax:
Based on https://github.com/mozilla/network-pulse/issues/1030: add a new field called `entry_type` to the Entry model, and update the get/post routes for entries to check for that value in preparation of adding model serializers that "fit" each of the entry types defined in our [table of entry types](https://docs.google.com/document/d/1LITsvpTJaafsyGuVrNeE27Sz_XY1Ase2D8IWCmu7dLk/edit?ts=5cadf65b#heading=h.p4wvs4fhqwpl)

gideonthomas:
@Pomax one other piece that would need to be done is how these entry types show up in the admin since the admin interface doesn't use our drf serializers. Probably should be a separate ticket?

Pomax:
pretty sure that as a charfield with a choice list, it'll just show up as a string in the admin interface?

Pomax:
![image](https://user-images.githubusercontent.com/177243/56612945-714a3a80-65ca-11e9-8009-bf43bdf4287b.png)

Pomax:
![image](https://user-images.githubusercontent.com/177243/56612970-7909df00-65ca-11e9-88cc-b7abe51c3fc3.png)

gideonthomas:
I would imagine that we don't want to show certain fields in the admin based on your choice of entry types? For e.g. I don't think we would want to show get involved or image for the session entry type in the admin? That might be a nice to have, but just thinking about other users of the admin who might be confused by which fields need to be filled out.

Pomax:
Yeah that feels like a nice to have rather than a hard requirement: ideally the admin would switch what it shows "on the fly" simply by you picking a diffferent entry type, and that is "a lot of work(tm)" to make happen because now we're talking a full entry payload but JS hiding/revealing based on a python serializer... that gets really complex really fast.

gideonthomas:
Ah, yeah I was thinking more of just a JS show/hide solution without worrying about the payload; at least that's how I did it before.

But yeah makes sense - def a nice to have.

Pomax:
filed it to make sure we don't lose that thought, though!

PR #483, opened 2019-04-23
add entry_type to Entry model

Pomax:
closes https://github.com/mozilla/network-pulse-api/issues/479 by adding an `entry_type` field for picking an appropriate serializer for any entry.

In the absence of an explicit type, this field defaults to `entry_type = "base"`


pulseapi/entries/models.py
```diff
@@ -73,8 +73,22 @@ class Entry(models.Model):
     """
     A pulse entry
     """
+    ENTRY_TYPES = (
+        ('base','base'),
```
gideonthomas: need spaces between tuple elements (this will also make travis pass)

Pomax: Ah, flake8...

In cases like these, feel free to use the "suggestion" feature for comments, and I'll just hit that "accept suggestion" button - that way you don't end up working on the PR, but code changes can still be effected without anyone having to go in and edit files.

PR #490, opened 2019-05-07
Multiple serializers for Entry based on type

### LINKED ISSUE #480: Add modelserializers for Entry types
Once https://github.com/mozilla/network-pulse-api/issues/479 has been completed, add modelserializers to each entry in our [table of entry types](https://docs.google.com/document/d/1LITsvpTJaafsyGuVrNeE27Sz_XY1Ase2D8IWCmu7dLk/edit?ts=5cadf65b#heading=h.p4wvs4fhqwpl):

- [x] base modelserializer
- [x] project entry
- [x] news entry
- [x] curriculum entry
- [x] info/research entry
- [ ] session entry

See also: https://github.com/mozilla/network-pulse/issues/1030

Pomax:
Closes #480 by creating separate serializers, with associated API endpoints, for entry types curriculum, info, news, project, and session


Pipfile
```diff
@@ -24,6 +24,7 @@ whitenoise = "*"
 "psycopg2-binary" = "*"
 django-allauth = "*"
 oauth2client = "*"
+httpie = "*"
```
Pomax: ignore this for now, it just makes API testing easier during dev

pulseapi/entries/serializers/curriculum.py
```diff
@@ -0,0 +1,26 @@
...
+        model = Entry
+        read_only_fields = fields = (
+            'id',
+            'title',
+            'content_url',
+            'thumbnail',
+            'is_bookmarked',
+        )
+        extra_kwargs = {
+            'url': {
```
gideonthomas: not sure I understand what this is for? `url` doesn't seem to be a field in the `Entry` model

pulseapi/profiles/serializers/profile_entries.py
```diff
@@ -4,7 +4,7 @@
 from pulseapi.creators.models import EntryCreator
 from pulseapi.entries.models import Entry
 from pulseapi.profiles.models import UserBookmarks
-from pulseapi.entries.serializers import (
+from pulseapi.entries.serializers.base import (
```
gideonthomas: Since `serializers` is now a directory, I would suggest doing what Chris did for profile serializers in terms of structuring modules (creating an `__init__.py` with an `__all__` property). That way we don't have to make these import changes no matter how we restructure the different modules in the future and can keep the import as `from .serializers import (...)`.

pulseapi/entries/views.py
```diff
@@ -502,3 +508,19 @@ def post(self, request, *args, **kwargs):
                 "post validation failed - {}".format(validation_result),
                 status=status.HTTP_400_BAD_REQUEST
             )
+
+
+class ProjectEntriesListView(EntriesListView):
```
gideonthomas: hmm, have you tested this? I'm not sure if the `get_serializer_class` method in the parent class will override the `serializer_class` here because I think it does that if it's in the same class.

pulseapi/entries/views.py
```diff
@@ -17,11 +17,16 @@
 from rest_framework.parsers import JSONParser
 
 from pulseapi.entries.models import Entry, ModerationState
-from pulseapi.entries.serializers import (
+from pulseapi.entries.serializers.base import (
     EntrySerializerWithV1Creators,
     EntrySerializerWithCreators,
     ModerationStateSerializer,
 )
+from .serializers.project import ProjectEntrySerializer
```
gideonthomas: if we switch to the module approach I mentioned below, we can simply do `from .serializers import ProjectEntrySerializer, NewsEntrySerializer, ...`

pulseapi/entries/serializers/project.py
```diff
@@ -0,0 +1,42 @@
...
+        - get involved url
+        - "is WIP?"
+        - help categories
+        - thumbnail image
+
+    extraneous
+        - custom text
+        - custom selector
+
+    This is, effectively, the same as a base entry.
```
gideonthomas: If this is going to be a read/write serializer, we probably should:
1. set `fields` to include the required and useful fields
2. explicitly declare overrides to the fields in `EntrySerializerWithCreators`. For example, `content_url` is not required here but is for regular entries (we'll need to change the model to reflect that btw, coz currently it's not nullable but it should be) -- so in the base serializer we would explicitly declare `content_url = serializers.URLField(required=True)` and in this serializer we would declare `content_url = serializers.URLField()`.

Pipfile
```diff
@@ -32,4 +33,4 @@ flake8 = "*"
 pgcli = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.7.3"
```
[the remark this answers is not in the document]
Pomax: I didn't see anything, but 3.6 is _really_ old now (December 2016... O_O), and I certainly don't have it installed anymore. I'll be happy to revert it to 3.6 so we can do that change separately, but I don't see it misbehaving so far =)

pulseapi/entries/serializers/curriculum.py
```diff
@@ -0,0 +1,26 @@
+from ..models import Entry
+from .news import EntrySerializerWithCreators
```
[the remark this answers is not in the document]
Pomax: hm, yeah. Why did runserver not crash on that O_o

pulseapi/entries/serializers/curriculum.py
```diff
@@ -0,0 +1,26 @@
...
+    """
+    Serializes an entry based on the fact that we
+    know that it is a curriculum item.
+    
+    This has the same requirements as news items.
+    """
+
+    class Meta:
+        model = Entry
+        read_only_fields = fields = (
```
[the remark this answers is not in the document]
Pomax: None, I copied the `Meta` from EntryBaseSerializer but I was going to ask about that - can I do with out? (so I suspect the answer is "yes" =D)

pulseapi/entries/serializers/curriculum.py
Pomax: Gah, that should have been `content_url`

pulseapi/entries/views.py
Pomax: we don't want it to, though - this is active on a different endpoint URL, so we always want this view to use the project serializer. Were you thinking about the v1 vs. v2 distinction? (do we need that if this only exists as of v2, given that v1 is basically "locked in the past"?)

pulseapi/entries/serializers/curriculum.py
gideonthomas: is there a reason it's in `extra_kwargs`?

pulseapi/entries/views.py
gideonthomas: both actually (I didn't think about versioning so glad you brought that up). And yeah you're right, we don't want it to override, but just checking that it actually doesn't override the `serializer_class` property because in:
```py
class SomeView(ListAPIView):
    serializer_class = A

    def get_serializer_class(self):
        return B
```
The serializer used is `B` because it always calls `get_serializer_class` and `ListAPIView` has a default implementation for `get_serializer_class` that returns the value of the property `serializer_class`.

So, when inheriting:
```py
class SomeOtherView(SomeView):
    serializer_class = C
```
since it always calls `get_serializer_class`, I would think it would go up the inheritance chain (`SomeOtherView`-->`SomeView`-->`ListAPIView`) and find an implementation of `get_serializer_class` in `SomeView` and execute that...so the serializer used would be `B` and not `C`.

I'm not 100% sure about that last bit which is why I'm asking if you tested to see if it works.

pulseapi/entries/serializers/__init__.py
```diff
@@ -0,0 +1,36 @@
+from .base import (
+    requiredField,
```
gideonthomas: needs to be removed from here and `__all__`

pulseapi/entries/serializers/curriculum.py
```diff
@@ -0,0 +1,16 @@
...
+    """
+    Serializes an entry based on the fact that we
+    know that it is a curriculum item.
+    
+    This has the same requirements as news items.
+    """
+
+    class Meta:
+        model = Entry
+        fields = EntryWithCreatorsBaseSerializer.Meta.fields
```
gideonthomas: I think we want `EntrySerializerWithCreators` for all Entry type models instead because it knows how to read/write creators. This means we probably don't need to override the `Meta` class here.

pulseapi/entries/serializers/info.py
```diff
@@ -0,0 +1,17 @@
...
+    Serializes an entry based on the fact that we
+    know that it is an info/research item.
+    
+    This has the same requirements as news items.
+    """
+
+    class Meta:
+        model = Entry
+        fields = EntryWithCreatorsBaseSerializer.Meta.fields
+
```
gideonthomas: extra line at the bottom of this file

pulseapi/entries/serializers/news.py
```diff
@@ -0,0 +1,37 @@
...
+from ..models import Entry
+from .base import EntryWithCreatorsBaseSerializer, EntrySerializerWithCreators
+
+
+class NewsEntrySerializer(EntrySerializerWithCreators):
+    """
+    Serializes an entry based on the fact that we
+    know that it is a news item. which means:
+
+    required:
```
gideonthomas: do we still want this comment?

pulseapi/entries/serializers/news.py
```diff
@@ -0,0 +1,37 @@
...
+        - description
+        - issue categories
+        - tags
+        - contributors
+        - get involved
+        - get involved url
+        - thumbnail image
+
+    extraneous
+        - help categories
```
gideonthomas: let's actually remove help categories from this serializer. I don't think it makes sense for news. In this case we need to override the Meta class like so (and preferably use `exclude` here instead of `fields`):
```py
class Meta(EntrySerializerWithCreators.Meta):
    exclude = EntrySerializerWithCreators.Meta.exclude + ('help_categories',)
```

pulseapi/entries/serializers/project.py
```diff
@@ -0,0 +1,40 @@
...
+        - custom text
+        - custom selector
+
+    This is, effectively, the same as a base entry,
+    but with the content_url left unrestricted.
+    """
+
+    content_url = serializers.URLField(required=False)
+
+    class Meta:
```
gideonthomas: same here, don't need to override `Meta`

pulseapi/entries/serializers/session.py
```diff
@@ -0,0 +1,33 @@
...
+    """
+    Serializes an entry based on the fact that we
+    know that it is a session item.
+    
+    This has the same requirements as project items.
+    """
+
+    content_url = serializers.URLField(required=False)
+
+    class Meta:
```
gideonthomas: probably should just be:
```py
class Meta(EntrySerializerWithCreators.Meta):
    exclude = EntrySerializerWithCreators.Meta.exclude + ('thumbnail', 'get_involved', 'get_involved_url',)
```

pulseapi/entries/tests/test_staff_views.py
```diff
@@ -11,7 +11,7 @@
 from pulseapi.profiles.models import UserProfile
 from pulseapi.entries.models import Entry, ModerationState
 from pulseapi.helptypes.factory import HelpTypeFactory
-from pulseapi.entries.serializers import (
+from pulseapi.entries.serializers.base import (
```
gideonthomas: we don't need this change anymore

pulseapi/entries/views.py
```diff
@@ -17,11 +17,19 @@
 from rest_framework.parsers import JSONParser
 
 from pulseapi.entries.models import Entry, ModerationState
-from pulseapi.entries.serializers import (
+from pulseapi.entries.serializers.base import (
```
gideonthomas: don't need this change, we can just add the serializers below to this list

pulseapi/profiles/serializers/profile_entries.py
gideonthomas: don't need this change (same with other files)

pulseapi/profiles/test_views.py
```diff
@@ -10,7 +10,7 @@
 
 from pulseapi.tests import PulseMemberTestCase
 from pulseapi.entries.models import Entry, ModerationState
-from pulseapi.entries.serializers import (
+from pulseapi.entries.serializers.base import (
```
gideonthomas: nit: we don't need `.base`

pulseapi/profiles/views/profile_entries.py
```diff
@@ -3,7 +3,7 @@
 from rest_framework.response import Response
 from rest_framework.views import APIView
 
-from pulseapi.entries.serializers import EntryWithCreatorsBaseSerializer, EntryWithV1CreatorsBaseSerializer
+from pulseapi.entries.serializers.base import EntryWithCreatorsBaseSerializer, EntryWithV1CreatorsBaseSerializer
```
gideonthomas: nit: we don't need `.base`

pulseapi/entries/models.py
```diff
@@ -87,10 +87,11 @@ class Entry(models.Model):
         max_length=20,
         choices=ENTRY_TYPES,
         default=ENTRY_TYPES[0][0],
-        null=False,
     )
     title = models.CharField(max_length=140)
-    content_url = models.URLField()
+    content_url = models.URLField(
+        blank=True,
```
Pomax:
```suggestion
```

PR #492, opened 2019-05-07
Create models for storing multiple profile/program types

### LINKED ISSUE #491: Create models to store multiple profile and program types for profiles
Related to https://github.com/mozilla/network-pulse/issues/1055

This creates the models for allowing us to associate profiles with multiple profile and program types with extra information like year, cohort, and order.

gideonthomas:
Fix #491 

This allows us to:
1. Associate profiles with multiple profile types and identify if they are current of past roles.
2. Associate profiles with multiple programs and identify which year/cohort they were associated with each program.

The relationships are ordered, e.g. it allows us to say a profile was a staff member and a fellow, in that order.


pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
...
+
+    year = models.PositiveSmallIntegerField(
+        # TODO: Change to MaxValueValidator with callable when
+        #       we update to Django > v2.2
+        validators=[YearValidator(max_offset=2)],
+        null=True,
+        blank=True,
+    )
+
+    cohort = models.CharField(
```
Pomax: Since "cohort" can be either a name or a year in natural language, it's probably worth calling this `cohort_name`

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
...
+
+    def clean(self):
+        # Don't allow both the cohort and year to be empty
+        if self.year is None and not self.cohort:
+            raise ValidationError(
+                _('Either the year or cohort must have a value')
+            )
+
+    class Meta:
+        verbose_name = 'Profile participation in a program'
```
Pomax: hm, what would the plural be?

pulseapi/utility/validators.py
```diff
@@ -0,0 +1,36 @@
...
+    message = _('The year must be between %(min_year)s and %(max_year)s.')
+    code = 'year_value'
+
+    def __init__(self, min_offset=10, max_offset=10):
+        self.min_offset = min_offset
+        self.max_offset = max_offset
+
+    def __call__(self, value):
+        current_year = date.today().year
+        min_year = current_year - self.min_offset
```
Pomax: This is a bit of a curious system: every year this would require we "bump-down" the min_offset just to preserve the min year. In this case it feels more sensible to peg the years, rather than to bracket around the current year.

pulseapi/profiles/models/profiles.py
```diff
@@ -182,6 +182,7 @@ def thumbnail_image_tag(self):
         default=False
     )
 
+    # TODO: Deprecate
```
Pomax: Just to make sure, this would be a followup with a migration function to build the new tuple objects?

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
 
     def __str__(self):
         return self.value
+
+
+class UserProfileProgramParticipation(models.Model):
```
Pomax: It feels like there should be a better name for this thing =D

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
...
+    program = models.ForeignKey(
+        'profiles.ProgramType',
+        on_delete=models.CASCADE,
+        related_name='participation',
+    )
+
+    year = models.PositiveSmallIntegerField(
+        # TODO: Change to MaxValueValidator with callable when
+        #       we update to Django > v2.2
+        validators=[YearValidator(max_offset=2)],
```
Pomax: This feels like something that would need yearly updating... (See comment below)

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
...
+class UserProfileProgramParticipation(models.Model):
+    profile = models.ForeignKey(
+        'profiles.UserProfile',
+        on_delete=models.CASCADE,
+        related_name='program_participation',
+    )
+
+    program = models.ForeignKey(
+        'profiles.ProgramType',
+        on_delete=models.CASCADE,
```
Pomax: I'm not sure CASCADE is the right policy here - cascading when the user gets deleted makes sense, but when a program gets deleted we probably want to preserve this "tuple" for historical "this user was in program X at some point, even though that program no longer exists"?

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,80 @@ class ProgramYear(models.Model):
...
+    profile_type = models.ForeignKey(
+        'profiles.ProfileType',
+        on_delete=models.CASCADE,
+        related_name='related_profiles',
+    )
+
+    is_current = models.BooleanField(default=True)
+
+    def __str__(self):
+        is_was = 'is' if self.is_current else 'was'
```
Pomax: best. varname.

pulseapi/profiles/models/categories.py
gideonthomas: ah good question I'll add that in as "List of profile/program associations" or something like that

pulseapi/utility/validators.py
gideonthomas: hmm, that's a very good point! Yeah I'll peg down the current year.

pulseapi/profiles/models/profiles.py
gideonthomas: I don't think it'll be a migration. I guess it's an open question - we can get rid of this at the model level OR, we can just keep it and use versioning to deprecate properties.

pulseapi/profiles/models/categories.py
gideonthomas: yeah I'm not good at naming bridge tables. Any suggestions?

pulseapi/profiles/models/categories.py
gideonthomas: Since the offset is from the current year we won't need to update. I'll peg down the min year (makes sense since it's static). But, for max_year, it will automatically change based on the current year. So every year, when we want to set a value for the `year` property, we can set it up to that year + 2 years ahead. So for example, in 2021, it will automatically allow to set the year to be up to 2023.

Now, the `2` offset is arbitrary (well sorta, I decided based on - we don't plan program members beyond 2 years). Happy to change that to something else.

pulseapi/profiles/models/categories.py
gideonthomas: I can change it. I was thinking that programs should never be deleted (even if it no longer exists).

pulseapi/profiles/models/categories.py
gideonthomas: I spent about 15 mins trying to figure out what word class "is" and "was" belong to. Is it a verb?

pulseapi/profiles/models/categories.py
Pomax: Thinking about it some more, this tuple basically literally encodes a `Cohort` record

pulseapi/profiles/models/categories.py
Pomax: if the program has no associated users, deleting should be fine (because it doesn't alter the public-facing data in any way) but between "not deleting programs" and "cascading a delete", the former feels like the better policy.

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,79 @@ class ProgramYear(models.Model):
...
+
+    program = models.ForeignKey(
+        'profiles.ProgramType',
+        on_delete=models.PROTECT,
+        related_name='profile_cohort_records',
+    )
+
+    year = models.PositiveSmallIntegerField(
+        # TODO: Change to MaxValueValidator with callable when
+        #       we update to Django > v2.2
```
Pomax: do we have an issue filed for updating Django? If not, let's file one and then we can refer to the gh link here, too.

pulseapi/profiles/models/categories.py
Pomax: 2 sounds reasonable to me.

pulseapi/profiles/models/categories.py
Pomax: it is. It's the present and past tense of "to be", a verb that acts as [copula](https://en.wikipedia.org/wiki/Copula_(linguistics)#Grammatical_function) in English.

pulseapi/profiles/models/categories.py
gideonthomas: https://github.com/MozillaFoundation/mofo-devops/issues/673

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,85 @@ class ProgramYear(models.Model):
...
+        self.full_clean()
+        super().save(*args, **kwargs)
+
+    class Meta:
+        verbose_name = 'cohort record'
+        # This meta option creates an _order column in the table
+        # See https://docs.djangoproject.com/en/1.11/ref/models/options/#order-with-respect-to for more details
+        order_with_respect_to = 'profile'
+        indexes = [
+            models.Index(fields=['profile', '_order'], name='uk_membership_profile_order'),
```
Pomax: what is `uk_`? (the only thing I can think of is the UK, which would be very strange =)

pulseapi/profiles/models/categories.py
```diff
@@ -59,3 +63,85 @@ class ProgramYear(models.Model):
...
+    def __str__(self):
+        copular_verb = 'is' if self.is_current else 'was'
+        return f'{self.profile.name} {copular_verb} a {self.role}'
+
+    class Meta:
+        # This meta option creates an _order column in the table
+        # See https://docs.djangoproject.com/en/1.11/ref/models/options/#order-with-respect-to for more details
+        order_with_respect_to = 'profile'
+        indexes = [
+            models.Index(fields=['profile', 'is_current', '_order'], name='uk_role_profile_current_order'),
```
Pomax: same here, what is `uk_`?

pulseapi/profiles/models/categories.py
gideonthomas: it's a db convention (so you'll only see this in the db) that stands for "unique key". We follow this for other indexes in pulse-api as well.

Pomax: I am having a hard time finding that mentioned anywhere in the django documentation - there isn't any mention of `uk_` anywhere on the link in line 114, so can I ask where this convention is codified, and can we add a comment with that link in one of the places `uk_` is used?

Pomax: after discussion on slack, R+ if there's a comment that explains what `uk_` means

PR #510, opened 2019-06-14
Create REMOVE_THIS.md

gideonthomas:
This is a temporary PR that will help us test the Join Us flow in the pulse repo.

### DIFF
--- a/REMOVE_THIS.md
+++ b/REMOVE_THIS.md
@@ -1,0 +1,1 @@
+REMOVE THIS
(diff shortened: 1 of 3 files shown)


REMOVE_THIS.md
```diff
@@ -0,0 +1 @@
+REMOVE THIS
```
Pomax: There is literally no need for this file if this PR also removes actual code

gideonthomas: The PR only contained this file initially, I didn't realize I needed other code changes until later.

Pomax: ah. worth taking this particular file back out then?

PR #512, opened 2019-06-20
auth: filter none values from google name

### LINKED ISSUE #511: Google auth sometimes returns None values
A user was unable to use GAuth to sign into Pulse (HTTP 500 server error).

From logentries:
```
web.1 - File "/app/pulseapi/users/adapter.py", line 88, in populate_user
web.1 - ]).strip() if not name else name
web.1 - TypeError: sequence item 1: expected str instance, NoneType found
```

Looks like Google auth can sometimes pass in `None` data (as opposed to omitting the key from the dict) to our auth system and this makes any operations on it fail.

gideonthomas:
Fix #511 

Not sure if there is a better way to do this in python.

### DIFF
--- a/pulseapi/users/adapter.py
+++ b/pulseapi/users/adapter.py
@@ -82,10 +82,12 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
     def populate_user(self, request, sociallogin, data):
         user = super().populate_user(request, sociallogin, data)
         name = data.get('name')
-        user.name = ' '.join([
-            data.get('first_name', ''),
-            data.get('last_name', '')
-        ]).strip() if not name else name
+        user.name = ' '.join(
+            list(filter(None, [
+                data.get('first_name', ''),
+                data.get('last_name', '')
+            ]))
+        ).strip() if not name else name
 
         if sociallogin.account.provider == google_provider_id and is_staff_address(user_email(user)):
             user.is_staff = True
--- a/pulseapi/users/adapter.py
+++ b/pulseapi/users/adapter.py
@@ -81,7 +81,7 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
     """
     def populate_user(self, request, sociallogin, data):
         user = super().populate_user(request, sociallogin, data)
-        
+
         name = data.get('name')
         if not name:
             first_name = data.get('first_name') or ''
--- a/pulseapi/users/adapter.py
+++ b/pulseapi/users/adapter.py
@@ -85,7 +85,7 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
         name = data.get('name')
         if not name:
             first_name = data.get('first_name') or ''
-            last_name = data.get('last_name')
+            last_name = data.get('last_name') or ''
             name = f"{first_name if first_name else ''} {last_name if last_name else ''}".strip()
         user.name = name if name else 'Unnamed Pulse user'
 
--- a/pulseapi/users/adapter.py
+++ b/pulseapi/users/adapter.py
@@ -86,7 +86,7 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
         if not name:
             first_name = data.get('first_name') or ''
             last_name = data.get('last_name') or ''
-            name = f"{first_name if first_name else ''} {last_name if last_name else ''}".strip()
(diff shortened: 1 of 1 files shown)


Pomax:
Can this "work"? Doesn't the `get('...', '')` syntax guarantee that we always get a value, so we'd never end up with a None in our list?

pulseapi/users/adapter.py
```diff
@@ -82,10 +82,12 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
     def populate_user(self, request, sociallogin, data):
         user = super().populate_user(request, sociallogin, data)
         name = data.get('name')
-        user.name = ' '.join([
-            data.get('first_name', ''),
-            data.get('last_name', '')
-        ]).strip() if not name else name
+        user.name = ' '.join(
```
Pomax: we probably want to shortcut this the other way:
```
user.name = name
if not user.name: 
  user.name = f'{data.get('first_name",'')} {data.get('last_name', '')}'.strip()
```

Pomax: we probably want to shortcut this the other way:
```
user.name = name
if not user.name: 
  user.name = f'{data.get('first_name",'')} {data.get('last_name', '')}'.strip()
```

Pomax: Although of course the notion that a name is composed of a first/last name at all is _exceptionally_ flawed, and reconstituting it with a space in between is equally and independently flawed. Works (often but not always) for western names, breaks very quickly outside of that cultural context =D

Pomax:
Do we know if it's possible to have no `name` _and_ no first/last name? If so we might still need a secondary fallback value

pulseapi/users/adapter.py
gideonthomas: the issue isn't about joining but that we get `None` values in the result so using an `f` string will not fix that, we need to filter it out.

gideonthomas:
> Do we know if it's possible to have no `name` _and_ no first/last name? If so we might still need a secondary fallback value

@Pomax Don't think we need that because there is a custom name value option. or if we do, let's do that separately because

pulseapi/users/adapter.py
Pomax: If `data.get('first_name', '')` can yield `None`, then we should avoid using `.get()`, as the idea is that if there's data, it's real data, otherwise it's an empty string.

If we can't set a policy on the super class that deletes keys with None values, so that `.get()` will do what it's supposed to, let's use a plain get without fallback value, and a formatted string that because it's breaking an expected pattern. Rather than filtering, can we use something like this?

```
        if not name:
            first = data.get('first_name')
            last = data.get('last_name')
            full_name =  f"{first if first else ''} {last if last else ''}".strip()
            user.name = full_name if full_name else ???
```

Where that `???` is kind of an open question: clearly it's _possible_ that there's no name, and no first/last name either, if `None` is an allowed data value, so we probably do want that ultimate fallback value in there (even if it's almost certainly never getting used)

pulseapi/users/adapter.py
```diff
@@ -81,11 +81,13 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
...
     def populate_user(self, request, sociallogin, data):
         user = super().populate_user(request, sociallogin, data)
+        
         name = data.get('name')
-        user.name = ' '.join([
-            data.get('first_name', ''),
-            data.get('last_name', '')
-        ]).strip() if not name else name
+        if not name:
+            first_name = data.get('first_name')
```
Pomax:
```suggestion
            first_name = data.get('first_name') or ''
```

pulseapi/users/adapter.py
```diff
@@ -81,11 +81,13 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
...
+            last_name = data.get('last_name')
```
Pomax:
```suggestion
            last_name = data.get('last_name') or ''
```

pulseapi/users/adapter.py
```diff
@@ -81,11 +81,13 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
...
+            name = f"{first_name if first_name else ''} {last_name if last_name else ''}".strip()
```
Pomax:
```suggestion
            name = f"{first_name} {last_name}".strip()
```

PR #515, opened 2019-06-28
fix the fact that with the regex version, local CORS does not work

Pomax:
When running network-pulse-api master and networl-pulse master together locally, I cannot for the life of me get them to cooperate, generating a million and one CORS errors in my browser dev console every time the pulse client tries to XHR some data from the pulse API... 

So this seems to be the only way I can make it work: remove the regex matching and stick with fully specified domains and ports =(

**if** we land this, we'll need to update our heroku env. vars to make sure the CORS whitelist has the correct values.

Testing for this is easy, but does require first checking out master for both pulse and pulse-api, running both, and verifying that there are CORS errors when loading http://localhost:3000 as well as http://test.example.com:3000 (if you have a test.example.com <=> 127.0.0.1 binding set up in your host file).

And yes, I added all of you for review mostly because I just want to make sure we all agree that CORS errors _actually_ occur. If one person can't reproduce it, there might be deeper magic at play =S


gideonthomas:
I'm testing this locally now to see if I can reproduce (having trouble running the front-end).
But, one problematic thing about this is that it's going to break review apps because we rely on regex for making review apps work

Pomax:
closing for now.

PR #516, opened 2019-06-28
created_at for profiles, with a default of jan 26, 2016

### LINKED ISSUE #514: Add "created" field to profile and user model
Currently, we don't record a create date on new profiles or new users. This fall, we aim to onboard program applicants more diligently (e.g. awards, mozfest). Create date will help us track whether our work is effective or not. Let's add it.

Pomax:
Closes #514 by adding a `created_at` property that defaults to Jan 26th, 2016, if it wasn't set before.


alanmoo:
@xmatthewx you wanted user accounts to have a creation date too, right?

xmatthewx:
yes. it's useful to measure growth and could be useful if we need to triage some spam or abuse.

pulseapi/profiles/migrations/0022_userprofile_created_at.py
```diff
@@ -0,0 +1,27 @@
...
+from __future__ import unicode_literals
+
+from django.db import migrations, models
+from django.utils.dateparse import parse_date
+from django.utils.timezone import make_aware
+
+def default_start_time():
+    date_str = "2016-01-26"
+    default_date = parse_date(date_str)
+    return make_aware(default_date)
```
gideonthomas: I think `default_date` needs to be a `datetime` object not a `date` object...that's probably why travis is failing as well.

You probably want to use something like `parse_datetime('2016-01-26 00:00:00')` as the default date instead and then you can make it timezone aware.

alanmoo: @Pomax Can you get this over the edge?

Pomax: yar

Pomax:
I've updated this to be two migrations: one that adds the field, and makes it default to jan 1st, 2016, and a second migration which then switches the field definition over to always be whatever "now" is when the profile gets created.

This needed two migrations because `default` and `auto_now_add` are mutually exclusive,  and we wanted a migration that did both.

pulseapi/profiles/migrations/0022_userprofile_created_at.py
```diff
@@ -0,0 +1,22 @@
...
+
+    dependencies = [
+        ('profiles', '0021_auto_20190508_1858'),
+    ]
+
+    operations = [
+        migrations.AddField(
+            model_name='userprofile',
+            name='created_at',
+            field=models.DateTimeField(default=datetime.datetime(2016, 1, 26, 0, 0, tzinfo=utc)),
```
Pomax: generated by `makemigration` based on:
```
    date_str = '2016-01-26 00:00:00'
    default_date = parse_datetime(date_str)
    timezoned = make_aware(default_date)
    created_at = models.DateTimeField(
        default=timezoned
    )
```

xmatthewx:
Good catch. I don't think it matters though.

Pomax:
Kept the field in the admin, but added it to the exclusion list for the public API

pulseapi/profiles/migrations/0023_auto_20190807_1733.py
```diff
@@ -0,0 +1,20 @@
...
+
+    dependencies = [
+        ('profiles', '0022_userprofile_created_at'),
+    ]
+
+    operations = [
+        migrations.AlterField(
+            model_name='userprofile',
+            name='created_at',
+            field=models.DateTimeField(auto_now_add=True),
```
Pomax: immediate second migration that changes the identity of `created_at` to one that can only ever be "whatever the date/time is right now" when a profile gets created.

Issue #544, opened 2019-09-30
uplift DRF to the latest version (epic)


Pomax:
After review, we need to do this in three parts:
- [x] uplift to 3.6.4
- [x] fix our mistake around filtering (see https://github.com/encode/django-rest-framework/pull/7094)
- [x] uplift to 3.10.x

Pomax:
Turns out our code allows upgrading all the way to 3.10.3

Pomax:
PR pending as https://github.com/mozilla/network-pulse-api/pull/546

Pomax:
https://github.com/mozilla/network-pulse-api/pull/545 broke the project

Pomax:
Problems and requirements: https://mozilla.slack.com/archives/C4EGUBY2Z/p1575666098003700

Pomax:
Copy-paste here, because who know show Slack actually keeps data:
```
<gideon> we need to back-out the last pulse-api prod release. It broke filtering, e.g. https://www.mozillapulse.org/projects?keyword=mozfest-2019%20bloggrid should show results and this affects the MozFest website -- see News section (maybe some fo'mo pages as well)
<pomax> I found a case that reproduces this bug in the test data, so I'm going to see if there is an easy fix, but if I don't find one in the next 15ish, let's roll back
<gideon> sounds good, thanks!
<pomax> right, this one's geniunely puzzling, and django is being its characteristic "lol you won't find a single fucking article that explains this unless you know the solution so you can use that to search the web with"
<alanmoo> Sounds like django
<pomax> so, I'll create a DRF branch off of current master, then roll master back to dec 3
<pomax> the problem appears to be that despite search_fields supposedly searching using "AND" rules if you give it more than one term, that is not what happens. for some combinations
<pomax> even in the test data, if I search for search=curriculum I get results, if I search for search=community,curriculum I get fewer, and correct, results, if I search for search=curriculum,libraries I get 0 results, despite there being plenty test entries with both tags.
<pomax> @gideon can you confirm things work now
<pomax> (the URLs at the start of this tread both work again as far as I can tell)
<pomax> https://www.mozillapulse.org/projects?keyword=mozfest-2019%20bloggrid also works
<pomax> ... now that I see https://www.mozillapulse.org/entry/1200.... are those hash marks the actual problem?
<gideon> seems to work now.
<gideon> I think we add the # on the front-end
<gideon> Thanks for the fix Pomax!
<pomax> make no mistake: this is not a fix =(
[...]

PR #545, opened 2019-09-30
DRF uplift

Pomax:
closes https://github.com/mozilla/network-pulse-api/issues/544 by uplifting from 3.5 to 3.10 by cycling through each intermediate version, with the associated code changes as indicated by https://www.django-rest-framework.org/community/3.5-announcement through https://www.django-rest-framework.org/community/3.5-announcement.


pulseapi/entries/serializers/news.py
```diff
@@ -2,6 +2,7 @@
 
 
 class NewsEntrySerializer(EntrySerializerWithCreators):
+    help_types = None
```
Pomax: as of DRF 3.7 fields may no longer be "active _and_ in the exclude list" so if they're inherited as active, they need to be explicitly set to `None` if they are to be excluded.

pulseapi/entries/views.py
```diff
@@ -248,7 +250,7 @@ def get_serializer_context(self):
     # When people POST to this route, we want to do some
     # custom validation involving CSRF and nonce validation,
     # so we intercept the POST handling a little.
-    @detail_route(methods=['post'])
+    @action(detail=True, methods=['post'])
```
Pomax: DRF 3.9 unified `detail_route` and `list_route` into a single `action` with a `detail` argument instead.

Pomax:
PRs really need a way to reopen when you revert them.

PR #555, opened 2019-10-20
docs: fix typo

Jason-Cooke:


Pomax:
a couple of months late, but: thanks for the fix!

PR #605, opened 2020-02-05
downgrade django-allauth to 0.40.0, which is the last version to support Django 1.x

Pomax:
Fixes https://github.com/mozilla/network-pulse-api/issues/604 by removing `pgcli` (fixing the psycopg2 issues) and pinning `django-allauth` to 0.40.0 (as 0.41.0 is the first version to no longer support Django 1.x). 

This also bumps up Django to 1.11.27 due to https://www.djangoproject.com/weblog/2019/dec/18/security-releases/


Pipfile
```diff
@@ -4,12 +4,14 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
+"psycopg2-binary" = "*"
 boto3 = "*"
 dj-database-url = "*"
-Django = "==1.11.23"
+Django = "==1.11.27"
```
patjouk: Can you bump it to 1.11.28 to cover the last scurity update? https://docs.djangoproject.com/en/dev/releases/1.11.28/

Pomax: I most certainly can.

Pomax:
updated, and review app settings updated as well.

PR #608, opened 2020-02-12
Use Python 3.7

cadecairos:
Update pipfile, travis, appveyor, and Heroku to use Python 3.7

Fixes #490

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -31,4 +31,4 @@ docutils = "*"
 flake8 = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.7"
--- a/.gitignore
+++ b/.gitignore
@@ -82,6 +82,7 @@ venv
 
 #Editor
 /.vscode/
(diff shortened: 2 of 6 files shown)


cadecairos:
clearing review while I figure out CI

cadecairos:
As of writing, Appveyor only supports up to python 3.7.5 (https://www.appveyor.com/docs/windows-images-software/#python)

Pipfile
```diff
@@ -31,4 +31,4 @@ docutils = "*"
 flake8 = "*"
 
 [requires]
-python_version = "3.6"
+python_version = "3.7"
```
cadecairos: we could specify the latest, but some distributions (I'm looking at _you_ ubuntu LTS) make it stupid difficult to install python 3.7, let alone 3.7.6

Pomax: I'm okay with major/minor only

PR #614, opened 2020-02-25
Uplift to DRF 3.6.4

Pomax:
With a fix for the kind of filtering we rely on based on https://stackoverflow.com/questions/60326973


pulseapi/entries/tests/test_member_views.py
```diff
@@ -147,4 +147,4 @@ def test_dual_tag_search(self):
         self.assertEqual(response.status_code, 200)
         response_data = json.loads(str(response.content, 'utf-8'))
         count = response_data['count']
-        self.assertEqual(count, 1)
+        self.assertEqual(count, 3)
```
Pomax: I'd like to bring this back down to 1, but so far extraneous matching is better than "no matching at all".

cadecairos: So is the change here that it previously matched entries tagges with "curricculum AND libraries" but now search matches "curricculum OR libraries OR (curricculum AND libraries"?

Pomax: how did you find this line? I changed this back to 1 over an hour ago O_o

Issue #615, opened 2020-02-26
⚠️ current master is broken ⚠️


Pomax:
https://network-pulse-api-staging.herokuapp.com/api/pulse/v2/entries/ is throwing `server error 500` despite CI passing.

Server log:

```bash
2020-02-26T05:51:39.658365+00:00 app[web.1]: Internal Server Error: /api/pulse/v2/entries/
2020-02-26T05:51:39.658410+00:00 app[web.1]: Traceback (most recent call last):
2020-02-26T05:51:39.658412+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/django/core/handlers/exception.py", line 41, in inner
2020-02-26T05:51:39.658412+00:00 app[web.1]:     response = get_response(request)
2020-02-26T05:51:39.658413+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/django/core/handlers/base.py", line 217, in _get_response
2020-02-26T05:51:39.658413+00:00 app[web.1]:     response = self.process_exception_by_middleware(e, request)
2020-02-26T05:51:39.658414+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/django/core/handlers/base.py", line 215, in _get_response
2020-02-26T05:51:39.658414+00:00 app[web.1]:     response = response.render()
2020-02-26T05:51:39.658414+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/django/template/response.py", line 107, in render
2020-02-26T05:51:39.658415+00:00 app[web.1]:     self.content = self.rendered_content
2020-02-26T05:51:39.658415+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/rest_framework/response.py", line 72, in rendered_content
2020-02-26T05:51:39.658416+00:00 app[web.1]:     ret = renderer.render(self.data, accepted_media_type, context)
2020-02-26T05:51:39.658416+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/rest_framework/renderers.py", line 703, in render
2020-02-26T05:51:39.658417+00:00 app[web.1]:     context = self.get_context(data, accepted_media_type, renderer_context)
2020-02-26T05:51:39.658417+00:00 app[web.1]:   File "/app/.heroku/python/lib/python3.7/site-packages/rest_framework/renderers.py", line 680, in get_context
[...]

PR #616, opened 2020-02-26
Fix missing templates

Pomax:
Fixes https://github.com/mozilla/network-pulse-api/issues/615

This fixes the missing templates errors due to uplifting DRF 3.5.3 to 3.6.3 - that PR forgot to add `django_filters` in the `INSTALLED_APPS` list. Oddly enough, testing passes fine, which suggests we're bypassing the entire "actually running the server" part of testing that everything works... =S

Test: check out bcf67ca05e99a7b76f8193e6fd780d28a647483a, then run `rm .env && inv setup && npm start` and verify that the "browse the api" option on `http://localhost:8000` breaks. Then check out this PR, run the same commands, and verify that things work again.


cadecairos:
> "actually running the server" part of testing that everything works

heh, unit testing != integration testing

cadecairos:
Tested & working for me locally.

Pomax:
"does the server come up and not crash python" is more of a unit test than an integration test tbh

PR #619, opened 2020-02-26
Django 2.0

Pomax:
closes https://github.com/mozilla/network-pulse-api/issues/620

It looks liike Django 2.0.13 is the current "dead end", because `2.1` and up change things such that our fake data causes integrity errors:

```bash
django.db.utils.IntegrityError: UNIQUE constraint failed: profiles_programtype.value
```

as well as mysteriously letting a route that was a 400 previously go through just fine:

```bash
Traceback (most recent call last):
  File "network-pulse-api\pulseapi\entries\tests\test_staff_views.py", line 479, in test_post_authentication_requirement
    self.assertEqual(postresponse.status_code, 400)
AssertionError: 200 != 400
```


Pomax:
the 4xx turning into a 2xx happens when going up from 2.0.13 to 2.1 - rebasing to fix that piplock

PR #624, opened 2020-03-10
Bump Django up to 2.1.15

Pomax:
This seems to fail a single test now, generating an HTTP200 success code, rather than HTTP400 error code, and that's rather puzzling.

### DIFF
--- a/pulseapi/entries/views.py
+++ b/pulseapi/entries/views.py
@@ -525,10 +525,9 @@ class EntriesListView(ListCreateAPIView):
     def post(self, request, *args, **kwargs):
         request_data = request.data
         user = request.user if hasattr(request, 'user') else None
-
         validation_result = post_validate(request)
 
-        if validation_result:
+        if validation_result is True:
             # invalidate the nonce, so this form cannot be
             # resubmitted with the current id
             request.session['nonce'] = False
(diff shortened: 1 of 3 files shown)


Pomax:
Haaaaa found it, and now I'm worried about how many posts we allowed that should not have >_>;;

pulseapi/entries/views.py
```diff
@@ -525,10 +525,9 @@ def get_serializer_context(self):
     def post(self, request, *args, **kwargs):
         request_data = request.data
         user = request.user if hasattr(request, 'user') else None
-
         validation_result = post_validate(request)
 
-        if validation_result:
+        if validation_result is True:
```
Pomax: @cadecairos so... yeah...  hmm.

cadecairos: /facepalm

cadecairos: "When JavaScript Developers start coding in Python"

Pomax: Oh, no, this would **absolutely** be the same bug in JS, because `validation_result` can be `True` _or string data_

Pomax: In fact, we had the correct check _in the same file, for the same var_ so I don't even know how this got here.

PR #628, opened 2020-03-27
Add size limit in admin

patjouk:
As I continue to work on optimizing the foundation site, I was surprised to see that images coming from pulse are well above 1MB.
For example, [the homepage](https://foundation.mozilla.org/en/) is 6MB, 4MB is images, 3.5MB **just** for pulse images.
Want something even scarier: https://www.mozillapulse.org/profile/82 is 61MB( image: 58MB), https://www.mozillapulse.org/latest is 14MB (images: 12MB)

Why 500kb? Because I was looking at the foundation site homepage and all the other images are below it and are still decent.

I'm adding this in admin to make sure that people login there to add entries won't be able to sneak something big again.

Similar PR is coming for pulse front.

### DIFF
--- a/pulseapi/entries/forms.py
+++ b/pulseapi/entries/forms.py
@@ -2,6 +2,7 @@ from django import forms
 from ajax_select.fields import AutoCompleteSelectMultipleField
 
 from pulseapi.creators.models import EntryCreator
+from pulseapi.utility.image_validation import enforce_image_size_limit
 
 
 class EntryAdminForm(forms.ModelForm):
@@ -42,8 +43,4 @@ class EntryAdminForm(forms.ModelForm):
 
     def clean_thumbnail(self):
         data = self.cleaned_data["thumbnail"]
-        if not data:
-            return data
-        if data.size > 512*1024:
-            raise forms.ValidationError("Image size must be less than 500KB")
-        return data
+        return enforce_image_size_limit(data)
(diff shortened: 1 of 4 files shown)


cadecairos:
In the review app admin, if I try to list profiles, I get a server timeout after 30 seconds - doesn't happen locally, or on production - @patjouk is this behaviour related to the PR somehow?

pulseapi/entries/forms.py
```diff
@@ -39,3 +39,11 @@ def _save_m2m(self):
         for creator_profile_id in creator_profile_id_list:
             profile_id = int(creator_profile_id)
             EntryCreator.objects.create(entry=entry, profile_id=profile_id)
+
+    def clean_thumbnail(self):
+        data = self.cleaned_data["thumbnail"]
+        if not data:
+            return data
```
Pomax: might be cleaner to say `if data and data.size > ...` on the next line, instead?

pulseapi/entries/forms.py
```diff
@@ -39,3 +39,11 @@ def _save_m2m(self):
         for creator_profile_id in creator_profile_id_list:
             profile_id = int(creator_profile_id)
             EntryCreator.objects.create(entry=entry, profile_id=profile_id)
+
+    def clean_thumbnail(self):
```
Pomax: As code used here as well as `pulseapi/profiles/forms.py`, lets put this in its own file in the utility dir, and the import it in both these forms.py files? 

E.g. something like:
```
from pulseapi.utility.form_helpers import disallow_large_images

...

    def clean_thumbnail(self):
        data = self.cleaned_data.get("thumbnail", None)
        return disallow_large_images(data) 
```

patjouk:
@cadecairos it's unrelated to this PR. I opened another one for optimizing the queries for the profile pages in the admin.

PR #642, opened 2020-05-18
remove pipenv

patjouk:
Remove pipenv and replace it by pip-tools.

Ref: mozillafoundation/mofo-devops#742


patjouk:
I moved the `tox.ini` file to the root of the project. That way, we don't have to pass it to flake8.

tasks.py
```diff
@@ -12,21 +12,98 @@
...
 
 
-@task(optional=['option', 'flag'])
-def manage(ctx, command, option=None, flag=None):
-    """Shorthand to manage.py. inv manage [COMMAND] [-o OPTION] [-f FLAG]. ex: inv manage runserver -o 3000"""
+@task
+def new_db(ctx):
+    """Create a new database with fake data"""
+    print("* Reset the database")
+    ctx.run("dropdb --if-exists pulse")
```
cadecairos, line 88: This command fails to connect to a database.

```
(pulsevenv)  cade@X1MOZCARBON  ~/code/network-pulse-api   remove-pipenv  inv new-db                                               
* Reset the database
dropdb --if-exists pulse
dropdb: could not connect to database template1: could not connect to server: No such file or directory
        Is the server running locally and accepting
        connections on Unix domain socket "/var/run/postgresql/.s.PGSQL.5432"?
```

cadecairos: started the postgres container with `docker run -e POSTGRES_HOST_AUTH_METHOD=trust -p 5432:5432  postgres`

patjouk:
Can you edit your `.env` to add those env vars:
```
DATABASE_URL=postgres://postgres@localhost:5432/pulse
ALLOWED_HOSTS=*
```

tasks.py
```diff
@@ -12,21 +12,98 @@
...
 
 
-@task(optional=['option', 'flag'])
-def manage(ctx, command, option=None, flag=None):
-    """Shorthand to manage.py. inv manage [COMMAND] [-o OPTION] [-f FLAG]. ex: inv manage runserver -o 3000"""
+@task
+def new_db(ctx):
+    """Create a new database with fake data"""
+    print("* Reset the database")
+    ctx.run("dropdb --if-exists -h localhost pulse")
```
cadecairos, line 88: This will also need a `-U postgres` otherwise it defaults to the username of the logged in user (in my case, `cade`)

cadecairos: We'll need this in the other places we interact with pg tools in this file.

patjouk: Good catch

PR #665, opened 2020-06-30
Invoke tasks support for both windows and posix systems

patjouk:
I forgot that windows and *Nix systems handle venv differently.
Not sure if it will work on the first try, but I don't have a windows system to test it. @Pomax can you check and edit this PR if necessary? Works fine on MacOS.

### DIFF
--- a/tasks.py
+++ b/tasks.py
@@ -116,7 +116,7 @@ def new_db(ctx):
 def manage(ctx, command):
     """Shorthand to manage.py. inv docker-manage \"[COMMAND] [ARG]\""""
     with ctx.cd(ROOT):
-        ctx.run(VENV_BIN_PATH + f"{PYTHON} manage.py {command}", **PLATFORM_ARG)
+        ctx.run(VENV_BIN_PATH + f"python manage.py {command}", **PLATFORM_ARG)
 
 
 @task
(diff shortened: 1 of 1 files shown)


tasks.py
```diff
@@ -5,48 +5,61 @@
...
+os.environ.pop("__PYVENV_LAUNCHER__", None)
 
 ROOT = os.path.dirname(os.path.realpath(__file__))
-
-# Python commands's outputs are not rendering properly. Setting pty for *Nix system and
-# "PYTHONUNBUFFERED" env var for Windows at True.
-if platform == 'win32':
-    PLATFORM_ARG = dict(env={'PYTHONUNBUFFERED': 'True'})
+VENV_DIR = "pulseenv"
+
```
Pomax, line 12: oops

```suggestion
```

tasks.py
```diff
@@ -85,8 +98,8 @@ def catch_up(ctx):
 def new_db(ctx):
     """Create a new database with fake data"""
     print("* Reset the database")
-    ctx.run("dropdb --if-exists -h localhost pulse -U postgres")
-    ctx.run("createdb -h localhost pulse -U postgres")
+    ctx.run("dropdb --if-exists -h localhost -U postgres pulse")
```
Pomax, line 101: got errors on Windows due to this ordering, which is weird, but also an easy fix

tasks.py
```diff
@@ -5,48 +5,59 @@
 
 # Workaround for homebrew installation of Python (https://bugs.python.org/issue22490)
 import os
-os.environ.pop('__PYVENV_LAUNCHER__', None)
+os.environ.pop("__PYVENV_LAUNCHER__", None)
```
Pomax, line 8: we were mixing `'` and `"` so a bunch of these changes are just cosmetic

patjouk: Thanks!

patjouk:
Reverting a change I made: we only need the `python`/`python3` for creating the venv. After that, we're calling `python` inside the venv.

Pomax:
A, good find - still works on my end!

Issue #762, opened 2022-01-24
Signup is getting abused for bitcoins and escort services. 


Pomax:
the pulse profile data has been poisoned through signup spam. 

- [x] disallow new signups
  - https://github.com/mozilla/network-pulse-api/pull/764
- [x] push to prod
- [x] set `ALLOW_SIGNUP` to False on prod
- [x] clean out all spam
  - About 1000 accounts got deleted. This will probably have included some legitimate users, but hopefully only a handful.
- [x] add recaptcha protection to the signup procedure:
  - [x] add server-side recaptcha validation for the `/login` route
    - [x] `USE_RECAPTCHA` and `RECAPTCHA_SECRET`
  - [x] add client-side recaptcha token generation in the navbar signin/signup link
    - [x] `USE_RECAPTCHA` and `RECAPTCHA_KEY`
    - https://github.com/mozilla/network-pulse/pull/1637
- [x] verify recaptcha works on staging
  - [x] merge API changes: https://github.com/mozilla/network-pulse-api/pull/766
  - [x] merge client changes: https://github.com/mozilla/network-pulse/pull/1637
  - [x] add `USE_RECAPTCHA` and `RECAPTCHA_KEY` to pulse pipeline's review app (through heroku)
  - [x] add `USE_RECAPTCHA` and `RECAPTCHA_SECRET` to pulse pipeline's staging instance (through heroku)
- [x] add prod env vars and push staging to prod 
- [x] add session lock to the API's `start_auth` (to only allow `accounts/login` if `/login` was hit first)
  - https://github.com/mozilla/network-pulse-api/pull/767
- [x] set `ALLOW_SIGNUP` back to True on prod
- [x] document the `ALLOW_SIGNUP` flag

PR #763, opened 2022-01-24
Disallow user signup

Pomax:
Part 1 of https://github.com/mozilla/network-pulse-api/issues/762


pulseapi/users/models.py
```diff
@@ -20,6 +26,7 @@ def create_user(self, name, email, password=None):
         if not password:
             password = self.make_random_password()
 
+
```
Pomax, line 29:
```suggestion
```

PR #766, opened 2022-01-25
Add recaptcha to the API server

Pomax:
This PR covers the server-side recaptcha work outlined in https://github.com/mozilla/network-pulse-api/issues/762

This gates the login route behind recaptcha validation, if `USE_RECAPTCHA` is enabled.

Requires https://github.com/mozilla/network-pulse/pull/1637 on the client-side.

# Testing

- clone the repo
- have python 3.9 installed - 3.10 is not going to work thanks to ancient dependencies. It is what it is.
- have pip and invoke installed
- run `inv setup`
- this will fail on the db creation part because it has no local postgres available.
- we don't care
- create an `.env` file and use the staging db url as `DATABASE_URL=...` because you do NOT want to set up a postgres 9.6 local instance, or even try to get docker to still do that with properly exposed ports.
- **DO NOT RUN `INV SETUP` OR `INV NEW-DB` FROM THIS POINT ON DURING TESTING**. You'll wipe the staging database. Please don't.
- add `USE_RECAPTCHA=True` to your `.env`
- grab the `localhost (v3)` recaptcha secret from https://www.google.com/recaptcha/admin and add `RECAPTCHA_SECRET=thatvalueyoujustgot` to your `.env`
- run `inv runserver`
- this... should work? If not, let me know, and let's figure out what's still missing.
- continue with the client instructions over on https://github.com/mozilla/network-pulse/pull/1637

### DIFF
--- a/pulseapi/settings.py
+++ b/pulseapi/settings.py
@@ -32,19 +32,26 @@ env = environ.Env(
     CORS_ORIGIN_REGEX_WHITELIST=(list, []),
     CORS_ORIGIN_WHITELIST=(list, []),
     CSRF_TRUSTED_ORIGINS=(list, []),
+    DATABASE_URL=(str, None),
     DEBUG=(bool, False),
     DJANGO_LOG_LEVEL=(str, 'INFO'),
-    DATABASE_URL=(str, None),
-    HEROKU_APP_NAME=(str, ''),
-    HEROKU_PR_NUMBER=(str, ''),
-    HEROKU_BRANCH=(str, ''),
-    PULSE_FRONTEND_HOSTNAME=(str, ''),
-    SECRET_KEY=(str, ''),
-    SSL_PROTECTION=(bool, False),
-    USE_S3=(bool, False),
     GITHUB_TOKEN=(str, ''),
+    HEROKU_APP_NAME=(str, ''),
+    HEROKU_BRANCH=(str, ''),
+    HEROKU_PR_NUMBER=(str, ''),
+    PULSE_FRONTEND_HOSTNAME=(str, ''),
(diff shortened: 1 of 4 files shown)


pulseapi/settings.py
```diff
@@ -32,20 +32,25 @@
     CORS_ORIGIN_REGEX_WHITELIST=(list, []),
     CORS_ORIGIN_WHITELIST=(list, []),
     CSRF_TRUSTED_ORIGINS=(list, []),
+    DATABASE_URL=(str, None),
```
Pomax, line 35: sorted alphabetically, only the `USE_RECAPTCHA` and `RECAPTCHA_SECRET` values are new.

pulseapi/users/utils.py
```diff
@@ -0,0 +1,68 @@
...
+RECAPTCHA_VERIFICATION_URL = 'https://www.google.com/recaptcha/api/siteverify'
+
+
+def verify_recaptcha(response_token):
+    """
+    Ask google to verify a client-generated recaptcha token
+    against what it supposedly generated.
+    """
+    try:
+        print(f"secret: {settings.RECAPTCHA_SECRET}")
```
Pomax, line 20: These print statements will be removed before landing, but you'll want them for debugging/testing.

danielfmiranda (approved):
Hi Pomax, just tested this out alongside the other PR and can confirm that it works! 
Also tested with the USE_RECAPTCHA variable turned off for one or the other, and for both, and am getting the expected results. 

Approving!

tbrlpld:
I have also tested in conjunction with the frontend PR and can confirm this works as expected.

tbrlpld (approved):

PR #767, opened 2022-01-27
Lock down the accounts/login route

Pomax:
This puts a session-based lock in front of the django-allauth `accounts/login` route, so that if recaptcha is enabled, users _have_ to first hit `/login` before they're redirected to `accounts/login`, yielding a 403 if they try to directly access the allauth login route.

### DIFF
--- a/pulseapi/users/utils.py
+++ b/pulseapi/users/utils.py
@@ -1,8 +1,10 @@
 import json
 
 from django.conf import settings
+from django.core.exceptions import PermissionDenied
 from django.http import HttpResponseForbidden
 from django.views.generic.base import RedirectView
+from allauth.account.views import LoginView
 
 from requests import post
 from requests.exceptions import RequestException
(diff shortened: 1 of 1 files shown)


pulseapi/users/utils.py
```diff
@@ -34,6 +36,43 @@ def verify_recaptcha(response_token):
...
+
+    if settings.USE_RECAPTCHA:
+        request = self.request
+        session = request.session
+
+        if 'recaptcha_token' not in session:
+            raise PermissionDenied()
+
+        session_token = session['recaptcha_token']
+        session['recaptcha_token'] = None
```
Pomax, line 59: clear the token after reading it so someone can't just set a valid session and then bulk-load the allauth route in a hundred tabs

tbrlpld: **Suggestion** This would probably make a good code comment  🙂

tbrlpld:
@Pomax One small suggestion. Looks good otherwise  👍

pulseapi/users/utils.py
Pomax, line 59:
```suggestion
        # clear the token after reading it so someone can't just set a valid session and then
        # bulk-load the allauth route with a hundred tabs.
        session['recaptcha_token'] = None
```

pulseapi/users/utils.py
```diff
@@ -34,6 +36,45 @@ def verify_recaptcha(response_token):
...
+
+    if settings.USE_RECAPTCHA:
+        request = self.request
+        session = request.session
+
+        if 'recaptcha_token' not in session:
+            raise PermissionDenied()
+
+        session_token = session['recaptcha_token']
+        # clear the token after reading it so someone can't just set a valid session and then
```
Pomax, line 59:
```suggestion
        session_token = session['recaptcha_token']

        # clear the token after reading it so someone can't just set a valid session and then
```

danielfmiranda (approved):
Hi @Pomax, thanks for tagging me in this with easy-to-follow comments! After reviewing the code and reading through the notes you left, everything looks good to me.

Approving! 👍

PR #769, opened 2022-02-03
reduce the autoreload logging, and allow custom runserver

Pomax:
no associated issue, but this makes it much better to work with the pulse API

### DIFF
--- a/pulseapi/settings.py
+++ b/pulseapi/settings.py
@@ -415,7 +415,7 @@ LOGGING = {
         },
         'django.utils.autoreload': {
             'handlers': ['debug'],
-           'level': 'INFO',
+            'level': 'INFO',
         }
     }
 }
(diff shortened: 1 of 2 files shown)


tasks.py
```diff
@@ -120,9 +120,9 @@ def manage(ctx, command):
 
 
 @task
-def runserver(ctx):
+def runserver(ctx, arguments):
     """Start a web server"""
-    manage(ctx, "runserver")
+    manage(ctx, f"runserver {arguments}")
```
Pomax, line 125: allows things like `inv runserver -- "0.0.0.0:8080"`

pulseapi/settings.py
```diff
@@ -413,6 +413,10 @@ def show_toolbar(request):
             'handlers': ['debug-error'],
             'level': 'ERROR'
         },
+        'django.utils.autoreload': {
+            'handlers': ['debug'],
+           'level': 'INFO',
```
Pomax, line 418: omg so much spam on the default `DEBUG` setting =_=

PR #775, opened 2022-02-08
Fix social adapter

Pomax:


tasks.py
```diff
@@ -111,6 +111,13 @@ def new_db(ctx):
           )
 
 
+# Django allauth
+@task(aliases=["secrets"])
+def generate_secrets(ctx):
+    with ctx.cd(ROOT):
+        ctx.run(VENV_BIN_PATH + f"python generate_client_secrets.py", **PLATFORM_ARG)
+
+
```
Pomax, line 120:
```suggestion
```

pulseapi/users/adapter.py
```diff
@@ -82,7 +82,7 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
         2. Handle the transition from the old auth system to "upgrade" existing
         accounts to the new allauth system.
     """
-    def is_open_for_signup(self, request):
+    def is_open_for_signup(self, request, socialaccount):
```
Pomax, line 85: as per https://github.com/pennersr/django-allauth/blob/198d356be0cc24081340b803624a2ae410c62155/docs/advanced.rst#creating-and-populating-user-instances

PR #788, opened 2022-04-06
Orphan profile cleanup + Removing profiles when user is deleted

### LINKED ISSUE #781: clean up orphan account profiles
it looks like profiiles are not set to cascade delete when a user account is deleted, so we have a bunch of orphan profiles hanging around that need cleaning up.

- [x] update the code so that a profile gets deleted rather than orphaned when a user's `email account` gets deleted
- [x] add a one-time "delete all orphaned profiles" management command that we can run to clean up the database.

danielfmiranda:
Closes #781 

This PR introduces a new "post_delete" signal for the EmailUser Model.


pulseapi/users/signals.py
```diff
@@ -1,14 +1,14 @@
 from django.dispatch import receiver
-from django.core.signals import post_save
+from django.db.models.signals import post_delete
 
 from pulseapi.users.models import EmailUser
 from pulseapi.profiles.models import UserProfile
 
 
-@receiver(post_save, sender=EmailUser)
```
danielfmiranda, line 8: It turns out that without setting up the app config in apps.py, this original "Create_profile_for_user" was never firing.

However when I updated the app config to pick this signal up, it brought the opposite of our original issue, orphan profiles were being created whenever a new user was created.

I think its safe to remove this, as I double checked that creating new users still creates a profile correctly without this signal. (the code for that can be found in users.models.py) and we can replace it with this post_delete signal which cleans up the profile after deleting a user.

PR #789, opened 2022-04-11
All the code changes to make sure new user accounts are marked as is_active=False

### LINKED ISSUE #779: make signups moderated
Allow signups to go through, but don't automatically mark new accounts as active. Instead, admins should flag accounts as active based on a knowledge of who the person behing the signup is.  That way we don't need to clamp down on signups as hard, and we can filter API resonses to only data associated with active user accounts.

- [x] allow signups, but leave new accounts "inactive"
- [ ] update the API routes for user and profile data to exclude inactive acccounts
- [ ] possibly update the API route for posts so that posts by deactivated users no longer show up?

Pomax:
closes #779 
left to do:
1. update the "inactivate account" template so users understand what's happening (this may need to be a follow-up separate issue)
1. make sure that the API route for profiles doesn't pull for any user with `is_active = False` (may need to be a follow-up separate issue rather than trying to tackle as part of account moderation)
1. make sure that the column name in the admin for the "is_active" flag has a shorter title (currently the same as the `help_text` which is pretty long)


danielfmiranda (approved):
Looks good to me and my update to the is active label has been added. Approved!

danielfmiranda:
New Issues for the 2 "Left to dos" 
1. [update the "inactivate account" template so users understand what's happening](https://github.com/mozilla/network-pulse-api/issues/790)
2.[ make sure that the API route for profiles doesn't pull for any user with `is_active = False`](https://github.com/mozilla/network-pulse-api/issues/791)

Issue #798, opened 2022-06-02
Update UserProfile.is_active to match EmailUser.is_active  


danielfmiranda:
In an effort to simplify and avoid future confusion, we would like to implement #791 and #795 using the `profile.is_active` flag instead of the `EmailUser.is_active` flag.

Our reasoning being:

1.If we move to the profile flag instead of the emailUser flag, It would allow users to sign up for pulse, verify their email, and edit their profiles/entries immediately, instead of being greeted with the "Inactive Account" page that was introduced earlier to combat spam accounts.
2. We could then use the profile `is_active` flag to filter out posts of "unapproved" users as requested in #791
3. Once we have these posts hidden, we can close #795, as the sign up flow will be back to the original flow.

cdanfon:
Hey @danielfmiranda Any updates on that pull request?

PR #799, opened 2022-06-05
updating profile.is_active flag based on user_account.is_active

danielfmiranda:
[**NOTE:**] We should coordinate when this can merge into the codebase, as this will allow "spam accounts" to sign up and to create posts without moderation 

Closes: #798 
Related PRs: #788, #789

In an effort to prepare for the tickets that will implement the new sign up flow:

1. New user logs in
2. They can access their profile and make entries
3. These posts and profile will be omitted from frontend until a admin updates their profile to "is_active".

We are now updating the app from "setting `user_account.is_active` to `False` on initial login" to "setting `profile.is_active` to `False` on initial login". That way the user can view/edit their profile, and we can then filter posts based on `profile.is_active`.

### DIFF
--- a/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
+++ b/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
@@ -10,14 +10,11 @@ def match_profile_isactive_to_user_isactive(apps, schema):
 
     for user in network_pulse_users:
 
-        users_profile, created = UserProfile.objects.get_or_create(related_user=user)
+        user_profile, created = UserProfile.objects.get_or_create(related_user=user)
 
-        if user.is_active:
-            users_profile.is_active = True
-        else:
-            users_profile.is_active = False
+        user_profile.is_active = user.is_active
 
-        users_profile.save()
+        user_profile.save()
 
 
 class Migration(migrations.Migration):
--- a/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
+++ b/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
@@ -24,7 +24,5 @@ class Migration(migrations.Migration):
     ]
 
     operations = [
-        migrations.RunPython(
-            match_profile_isactive_to_user_isactive
-        ),
+        migrations.RunPython(code=match_profile_isactive_to_user_isactive, reverse_code=migrations.RunPython.noop)
     ]
--- a/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
+++ b/pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
@@ -1,0 +1,33 @@
+from django.db import migrations
+from django.conf import settings
+
+
+def match_profile_isactive_to_user_isactive(apps, schema):
+    EmailUser = apps.get_model('users', 'EmailUser')
+    UserProfile = apps.get_model('profiles', 'UserProfile')
+
+    network_pulse_users = EmailUser.objects.all()
+
+    for user in network_pulse_users:
+
+        users_profile, created = UserProfile.objects.get_or_create(related_user=user)
+
+        if user.is_active:
+            users_profile.is_active = True
+        else:
+            users_profile.is_active = False
+
+        users_profile.save()
+
+
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('profiles', '0024_delete_orphan_profiles'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            match_profile_isactive_to_user_isactive
+        ),
+    ]
--- a/pulseapi/users/adapter.py
+++ b/pulseapi/users/adapter.py
@@ -111,7 +111,7 @@ class PulseSocialAccountAdapter(DefaultSocialAccountAdapter):
         try:
             UserProfile.objects.get(related_user=user)
         except UserProfile.DoesNotExist:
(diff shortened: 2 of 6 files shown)


pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
```diff
@@ -0,0 +1,33 @@
...
+
+def match_profile_isactive_to_user_isactive(apps, schema):
+    EmailUser = apps.get_model('users', 'EmailUser')
+    UserProfile = apps.get_model('profiles', 'UserProfile')
+
+    network_pulse_users = EmailUser.objects.all()
+
+    for user in network_pulse_users:
+
+        users_profile, created = UserProfile.objects.get_or_create(related_user=user)
```
danielfmiranda, line 13: Since there are still around 4 user accounts without linked profiles in the DB (bug we are aware of and going to create a ticket to investigate), I needed to use the "get_or_create" method so the migration doesnt return an error when trying to find a users profile.

pulseapi/users/signals.py
```diff
@@ -1,21 +1,12 @@
...
-    if instance._state.adding is True:
-        instance.is_active = False
-
-
 @receiver(post_delete, sender=EmailUser)
 def delete_profile_for_user(sender, **kwargs):
-    related_profile_id = kwargs['instance'].profile_id
-    related_profile = UserProfile.objects.get(id=related_profile_id)
-    related_profile.delete()
+    if kwargs['instance'].profile_id:
```
danielfmiranda, line 9: This if statement is to fix a bug that is occurring whenever I am trying to delete a account without a linked profile. 

Creating a ticket to investigate what is causing these "orphan profiles" in the first place.

pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
```diff
@@ -0,0 +1,33 @@
...
+
+        if user.is_active:
+            users_profile.is_active = True
+        else:
+            users_profile.is_active = False
```
tbrlpld, line 18: **Suggestion** You should be able to simplify this like so:

```suggestion
        user_profile.is_active = user.is_active
```

pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
tbrlpld, line 13: **Typo** I believe there is a typo in this line.
```suggestion
        user_profile, created = UserProfile.objects.get_or_create(related_user=user)
```

pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
```diff
@@ -0,0 +1,33 @@
...
+
+        users_profile.save()
```
tbrlpld, line 20: **Typo** I believe there is a typo in this line.
```suggestion
        user_profile.save()
```

pulseapi/users/signals.py
tbrlpld: **Question** Is that really an issue with "orphan profiles"? It sounds more like users without profiles.

tbrlpld: Maybe these users are created in an alternative paths that is not using the `EmailUserManage.create_user` method?

pulseapi/users/signals.py
```diff
@@ -1,21 +1,13 @@
...
+        related_profile_id = kwargs['instance'].profile_id
+        related_profile = UserProfile.objects.get(id=related_profile_id)
+        related_profile.delete()
```
tbrlpld, line 13: **Suggestion** To also protect against cases where the user might still have the profile ID set, but the profile does not really exist anymore (for whatever reason), you could also protect this line with an if statement. 
```suggestion
        if related_profile:
            related_profile.delete()
```

tbrlpld (changes requested):
This is looking good. Thanks @danielfmiranda. Just some minor questions and suggestions.

pulseapi/profiles/migrations/0025_transfer_user_isactive_to_profile.py
```diff
@@ -0,0 +1,33 @@
...
+class Migration(migrations.Migration):
+
+    dependencies = [
+        ('profiles', '0024_delete_orphan_profiles'),
+    ]
+
+    operations = [
+        migrations.RunPython(
+            match_profile_isactive_to_user_isactive
+        ),
```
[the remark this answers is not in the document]
danielfmiranda: Great catch Tibor, and thanks for sending me the docs! 

I have now updated the line to include the `noop` to prevent any problems during a backwards migration.

pulseapi/users/adapter.py
```diff
@@ -111,7 +111,7 @@ def save_user(self, request, sociallogin, form=None):
         try:
             UserProfile.objects.get(related_user=user)
         except UserProfile.DoesNotExist:
-            profile = UserProfile.objects.create(is_active=True)
+            profile = UserProfile.objects.create(is_active=False)
```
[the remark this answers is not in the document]
danielfmiranda: Comments have been added!

danielfmiranda:
Hi Tibor! Thanks for the great suggestions, I have implemented your feedback, re-requested review, and have created https://github.com/mozilla/network-pulse-api/issues/800 to investigate the user accounts with missing linked profiles.

tbrlpld (approved):
Thanks @danielfmiranda. This looks good to me 👍

tbrlpld (approved):

PR #801, opened 2022-06-12
updated API endpoints to omit inactive profiles and their entries

### LINKED ISSUE #791: Update API routes to exclude inactive accounts
After PR https://github.com/mozilla/network-pulse-api/pull/789 has been merged in, users who create an account will not be marked as active until a staff member logs in and grants them access. 

We should update the API routes so that they do not return any users/profiles/posts from users who are inactive.

danielfmiranda:
Closes #791 

This PR introduces the following changes to these endpoints:

/creators/:
--
 - Now using the `UserProfile` models [.active()](https://github.com/mozilla/network-pulse-api/blob/83b134e21526d0695106a49accdcef4b41384d01/pulseapi/profiles/models/profiles.py#L29) filter, to return only profiles with `'is_active'` set to `True` in the queryset.
 - Added a new test in [creators/test.py](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/creators/tests.py#L54) to verify that the filtering of active profiles works correctly.

/profiles/:
---
- Updated the queryset in [/views/profiles.py](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/profiles/views/profiles.py#L249) to include an `if` statement that checks whether or not someone is using the "`?is_active`" search param. This is so we retain functionality of this filter, and we can filter for "inactive" profiles through the API if needed. However, all other filters will default to  filtering out inactive accounts like requested.
- Updated [test_profile_list_filtering](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/profiles/test_views.py#L435) to set the test profiles `is_active` flag to `True`, as without this, they would be set to inactive and the view would return an empty queryset.

/entries/:
---
- Created a new filter for the entry queryset, called [by_active_profile()](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/entries/models.py#L71)
- Updated the public [EntriesList queryset](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/entries/views.py#L469) to return only entries by active profiles.
- Added a new test in [test_member_views.py](https://github.com/mozilla/network-pulse-api/blob/bb94b937eab123c057d129ce137ec1eeec36a6d5/pulseapi/entries/tests/test_member_views.py#L159) to verify that entries created by inactive users are hidden from the public view.

### DIFF
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -53,15 +53,20 @@ class TestEntryCreatorViews(PulseStaffTestCase):
 
     def test_hiding_inactive_profiles(self):
         """Testing that inactive profiles are not returned in the creator view"""
-        # This profiles "is_active" is False by default, and should not be returned
-        # in the API response.
         profile = UserProfile.objects.last()
+        profile.is_active = True
+        profile.save()
 
-        response = self.client.get('{creator_url}?name={search}'.format(
-            creator_url=reverse('creators-list'),
-            search=quote(profile.name)
+        inactive_profile = UserProfile.objects.first()
+        inactive_profile.is_active = False
+        inactive_profile.save()
+
+        response = self.client.get('{creator_url}'.format(
+            creator_url=reverse('creators-list')
         ))
         search_results = json.loads(str(response.content, 'utf-8'))['results']
+        list_of_profile_ids = [profile['profile_id'] for profile in search_results]
 
         self.assertEqual(response.status_code, 200)
-        self.assertEqual(search_results, [])
+        self.assertIn(profile.id, list_of_profile_ids)
+        self.assertNotIn(inactive_profile.id, list_of_profile_ids)
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -18,7 +18,7 @@ class TestEntryCreatorViews(PulseStaffTestCase):
         page = CreatorsPagination()
         expected_data = CreatorSerializer(
             page.paginate_queryset(
-                UserProfile.objects.filter(is_active=True).order_by('id'), # Filtering for active profiles only
+                UserProfile.objects.active().order_by('id'),  # Filtering for active profiles only
                 request=Request(request=HttpRequest())  # mock request to satisfy the required arguments
             ),
             many=True
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -18,7 +18,7 @@ class TestEntryCreatorViews(PulseStaffTestCase):
         page = CreatorsPagination()
         expected_data = CreatorSerializer(
             page.paginate_queryset(
-                UserProfile.objects.all().order_by('id'),
+                UserProfile.objects.filter(is_active=True).order_by('id'), # Filtering for only active profiles
                 request=Request(request=HttpRequest())  # mock request to satisfy the required arguments
             ),
             many=True
@@ -37,6 +37,10 @@ class TestEntryCreatorViews(PulseStaffTestCase):
     def test_creator_filtering(self):
         """search creators, for autocomplete"""
         profile = UserProfile.objects.last()
+        # Setting this profile's "is_active" flag to true, as the API is
+        # set up to only return active profiles.
+        profile.is_active = True
+        profile.save()
 
         response = self.client.get('{creator_url}?name={search}'.format(
             creator_url=reverse('creators-list'),
@@ -46,3 +50,18 @@ class TestEntryCreatorViews(PulseStaffTestCase):
 
         self.assertEqual(response.status_code, 200)
         self.assertEqual(profile.id, response_creator['creator_id'])
+
+    def test_hiding_inactive_profiles(self):
+        """Testing that inactive profiles are not returned in the creator view"""
+        # This profiles "is_active" is False by default, and should not be returned
+        # in the API response.
+        profile = UserProfile.objects.last()
+
+        response = self.client.get('{creator_url}?name={search}'.format(
+            creator_url=reverse('creators-list'),
+            search=quote(profile.name)
+        ))
+        search_results = json.loads(str(response.content, 'utf-8'))['results']
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(search_results, [])
--- a/pulseapi/creators/tests.py
+++ b/pulseapi/creators/tests.py
@@ -18,7 +18,7 @@ class TestEntryCreatorViews(PulseStaffTestCase):
         page = CreatorsPagination()
         expected_data = CreatorSerializer(
             page.paginate_queryset(
-                UserProfile.objects.filter(is_active=True).order_by('id'), # Filtering for only active profiles
+                UserProfile.objects.filter(is_active=True).order_by('id'), # Filtering for active profiles only
                 request=Request(request=HttpRequest())  # mock request to satisfy the required arguments
             ),
             many=True
--- a/pulseapi/entries/models.py
+++ b/pulseapi/entries/models.py
@@ -68,6 +68,12 @@ class EntryQuerySet(models.query.QuerySet):
             'related_entry_creators__profile__related_user',
         )
 
+    def by_active_profile(self):
+        """
+        Return all entries that have been created by pulse users who's profiles are set to "Active" by moderator.
+        """
+        return self.filter(published_by__profile__is_active=True)
+
 
 class Entry(models.Model):
     """
--- a/pulseapi/entries/tests/test_member_views.py
+++ b/pulseapi/entries/tests/test_member_views.py
@@ -5,6 +5,7 @@ from pulseapi.entries.models import Entry, ModerationState
 from pulseapi.creators.models import EntryCreator
 from pulseapi.tags.models import Tag
 from pulseapi.tests import PulseMemberTestCase
+from pulseapi.users.factory import BasicEmailUserFactory
 
 
 class TestMemberEntryView(PulseMemberTestCase):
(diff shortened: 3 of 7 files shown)


pulseapi/profiles/views/profiles.py
```diff
@@ -242,7 +242,14 @@ class UserProfileListAPIView(ListAPIView):
 
     def get_queryset(self):
         request = self.request
-        queryset = UserProfile.objects.all().prefetch_related('related_user')
+        queries = self.request.query_params
```
danielfmiranda, line 245: I noticed that we had a filter set up for the "is_active" flag, in an effort to keep this usable, I added this if statement.

That way, if a dev wants to visit the API and filter for all inactive users, they can still use the "?Is_active=False" search param.

Any other filtering however, will filter out inactive profiles by default.

pulseapi/profiles/test_views.py
```diff
@@ -478,8 +479,6 @@ def test_profile_list_filtering_active_v2(self):
         self.run_test_profile_list(
             api_version=settings.API_VERSIONS['version_2'],
             profile_serializer_class=UserProfilePublicSerializer,
-            profile_params={'is_active': True},
```
danielfmiranda, line 481: Since we are already filtering the QS to return only active profiles, I believe we can remove these!

pulseapi/entries/models.py
```diff
@@ -68,6 +68,12 @@ def with_related(self):
             'related_entry_creators__profile__related_user',
         )
 
+    def by_active_profile(self):
+        """
+        Return all entries that have been created by pulse users who's profiles are set to "Active" by moderator.
+        """
+        return self.filter(published_by__profile__is_active=True)
+
```
tbrlpld, line 76: **Question** There already is the `public` method on the `EntryQuerySet`. From the docstring and the filter it appears to serve the same purpose (only show moderator approved content). It seems like anything that we would want to show publically should fulfill both, the "approved" moderation state and that the user profile is "active". Is there need to split these checks? Or could we include the "profile is active" check in the `public` method?

pulseapi/entries/tests/test_member_views.py
```diff
@@ -148,3 +155,10 @@ def test_dual_tag_search(self):
         response_data = json.loads(str(response.content, 'utf-8'))
         count = response_data['count']
         self.assertEqual(count, 1)
+
+    def test_inactive_profile_entry_is_hidden(self):
+        response = self.client.get('/api/pulse/entries/?search=inactive_profile_entry&format=json')
+        self.assertEqual(response.status_code, 200)
+        response_data = json.loads(str(response.content, 'utf-8'))
+        count = response_data['count']
+        self.assertEqual(count, 0)
```
tbrlpld, line 164: **Suggestion** Similar to the other point, when we test exclusion, I think it makes sense to include another object that needs to be present as a sanity check to make sure our tests are actually working.

pulseapi/profiles/test_views.py
```diff
@@ -432,6 +432,7 @@ def test_profile_list_filtering(self):
                     (program_year, _) = ProgramYear.objects.get_or_create(value=v3)
                     profile = UserProfile.objects.create()
                     profile.enable_extended_information = True
+                    profile.is_active = True
```
tbrlpld, line 435: Wow... what a complex test... not because of your change of course 😅

pulseapi/profiles/views/profiles.py
tbrlpld: **Suggestion** It might make sense to add a test for this. Currently I believe this would return all profiles if `is_active` is part of the URL. E.g. `?is_active=True` and `?is_active=False` would return the same results. I think that's the case be cause the if-condition only checks if the parameter is in the included in the querystring, but not what its value is.

tbrlpld (changes requested):
@danielfmiranda I have left a few questions for clarification and suggestion.

pulseapi/creators/tests.py
```diff
@@ -46,3 +50,18 @@ def test_creator_filtering(self):
...
+        profile = UserProfile.objects.last()
+
+        response = self.client.get('{creator_url}?name={search}'.format(
+            creator_url=reverse('creators-list'),
+            search=quote(profile.name)
+        ))
+        search_results = json.loads(str(response.content, 'utf-8'))['results']
+
+        self.assertEqual(response.status_code, 200)
+        self.assertEqual(search_results, [])
```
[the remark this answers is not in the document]
danielfmiranda: Added test to make sure that inactive users are excluded while active users are included!

pulseapi/entries/tests/test_member_views.py
danielfmiranda: Updated!

pulseapi/profiles/views/profiles.py
danielfmiranda: Good Idea Tibor, I have added a new test where we:

- Create a test inactive profile to be sure there is one in the database.
- Preform a get request at the `profiles/?is_active` endpoint.
- Check if any of the returned profiles have `is_active` set to false.

pulseapi/entries/models.py
danielfmiranda: Hi Tibor!

The public() filter is used for two querysets: the `EntryQuerySet` which is used to show individual entries and the `EntryListQuerySet` which is used to display a list of all public entries by approved profiles.

The EntryQuerySet is called when a user clicks on an entry from their profile, regardless of whether they are set to `is_active` or not. If we were to combine these two filter into one, these inactive users would no longer be able to visit their posts until they are set to active which is not the desired workflow.

Let me know if that helps, thanks!

Pomax (approved):

pulseapi/entries/models.py
tbrlpld: That makes sense. Thanks @danielfmiranda. If this is for users to see their own posts, we could potentially also allow them to see posts that are not approved. Or am I missing something?

pulseapi/profiles/test_views.py
```diff
@@ -504,3 +504,20 @@ def test_profile_categories(self):
         self.assertListEqual(categories['profile_types'], list(ProfileType.objects.values_list('value', flat=True)))
         self.assertListEqual(categories['program_types'], list(ProgramType.objects.values_list('value', flat=True)))
         self.assertListEqual(categories['program_years'], list(ProgramYear.objects.values_list('value', flat=True)))
+
+    def test_profile_view_is_active_filter(self):
```
tbrlpld, line 508: **Suggestion** Similar to the other tests, it would be good to also create a sanity check object that is `is_active=True` and is included in the response.

pulseapi/profiles/test_views.py
```diff
@@ -504,3 +504,20 @@ def test_profile_categories(self):
...
+        """
+        Creating and saving an inactive profile, then preforming a GET
+        request to the profile endpoint with the '?is_active' filter set to
+        'True', then checking if any returned profiles are set to 'False'.
+        """
+        inactive_profile = BasicUserProfileFactory.create(is_active=False)
+        inactive_profile.save()
+
+        profile_url = reverse('profile_list')
+        url = ('{url}?is_active=True').format(url=profile_url)
```
tbrlpld, line 518: **Suggestion** I think it would be great to add another test where you use `url = ('{url}?is_active=False').format(url=profile_url)` and check that only the inactive profiles are returned. Because the "only active profiles" case is already the default, or am I off here?

tbrlpld (changes requested):
Nice @danielfmiranda. Thanks for updating the tests! These look a lot clearer to me now 🙌

pulseapi/profiles/test_views.py
danielfmiranda: Added [test_profile_view_is_inactive_filter](https://github.com/mozilla/network-pulse-api/blob/ad7ee392ea97c16a6febe6894082ae538c1b7c86/pulseapi/profiles/test_views.py#L527) 👍

pulseapi/profiles/test_views.py
danielfmiranda: Updated!

pulseapi/profiles/test_views.py
```diff
@@ -521,3 +523,22 @@ def test_profile_view_is_active_filter(self):
         num_of_inactive_profiles = sum(profile['is_active'] is False for profile in entriesjson)
 
         self.assertEqual(num_of_inactive_profiles, 0)
```
tbrlpld, line 525: **Suggestion** We should add an assertion to check that the active profile is returned here (or that the number of active profiles is 1). That is the data that we are requesting.

pulseapi/profiles/test_views.py
```diff
@@ -521,3 +523,22 @@ def test_profile_view_is_active_filter(self):
...
+        inactive_profile = BasicUserProfileFactory.create(is_active=False)
+        inactive_profile.save()
+
+        profile_url = reverse('profile_list')
+        url = ('{url}?is_active=False').format(url=profile_url)
+        response = self.client.get(url)
+        entriesjson = json.loads(str(response.content, 'utf-8'))
+        num_of_active_profiles = sum(profile['is_active'] is True for profile in entriesjson)
+
+        self.assertEqual(num_of_active_profiles, 0)
```
tbrlpld, line 544: **Suggestion** Here we should also add an assertion checking that the inactive profile is returned (or that their count is 1).

danielfmiranda:
Thanks for the suggestions @tbrlpld!
Per our discussion in slack, it was a little difficult to check for one active profile, as the testing DB is already populated with some other profiles.

What I implemented instead, is taking a list of all returned 'profile_ids' and asserting that the expected profile can be found in that list, while also asserting that the profile meant to be filtered out, is in fact left out of that list.

I have pushed these changes and re-requested review, please let me know if you have any more questions. Thanks again!

tbrlpld (approved):

PR #815, opened 2025-04-04
Move build dependencies into requirements.txt for Heroku compatibility

mmmavis:
This PR resolves [build issues on Heroku](https://dashboard.heroku.com/apps/network-pulse-api-staging/activity/builds/3caac6e0-fab2-4338-9dbf-021e750bc865) by moving the `setuptools` and `wheel` dependencies from `build-requirements.txt` into the main `requirements.txt` file.

Heroku uses `requirements.txt` as the default install target during the slug build process. Having `setuptools` and `wheel` there ensures that packages like `django-allauth`, which rely on older setup.py behavior, install correctly without metadata generation errors.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,9 +5,11 @@
 #    pip-compile
 #
 
-# Required on Heroku to prevent build failures with legacy packages like django-allauth
-# Ensures build dependencies (e.g. setuptools, wheel) are installed before other packages
--r build-requirements.txt
+# The two lines below are needed for Heroku compatibility.
+# `setuptools==65.5.1` ensures compatibility with legacy packages like django-allauth,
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -5,9 +5,11 @@
 #    pip-compile
 #
 
-# Required on Heroku to prevent build failures with legacy packages like django-allauth
-# Ensures build dependencies (e.g. setuptools, wheel) are installed before other packages
--r build-requirements.txt
+# The two lines below are needed for Heroku compatibility.
```
danielfmiranda, line 8: **Kudos** Thank you for updating the comment here!

danielfmiranda (approved):
Looks good to me! Thanks @mmmavis 👍 approved!

